PR #4, opened 2017-12-13
Add initial Travis CI curation file integration

mynameistechno:
First stab at build script that validates yaml files. Adds initial travis CI config as well.


mynameistechno:
Not sure if we can hide all the other log output but I was able to pretty print yaml and add some colors too make it pop out more. Failures will be red text.

image

PR #22, opened 2017-12-28
Update redie.yaml

jeffmcaffer:


fossygirl:
@capfei Please sanity check this @jeffmcaffer change otherwise we'll close if it's not good :-)

capfei:
It looks like a test submission. Closing again.

PR #93, opened 2018-08-24
AE49_180824_232949.367

AE49:
Add MIT (https://github.com/mono/guiunit/blob/master/LICENSE.txt)


capfei:
@AE49 The package does not have a declared a license and I cannot find one in the source. Can you point me to where it is?

fossygirl:
This is under a Microsoft EULA, closing.

PR #144, opened 2018-08-25
AE49_180825_092849.875

AE49:
Add MIT (https://github.com/Microsoft/ManagedEsent/blob/master/LICENSE.md)


capfei:
The NuGet page states MS-PL. Can you point me to where it states MIT for this version? I'm not seeing it.

AE49:
The license link on NuGet says "MS-PL" but when you click on it, there is no MS-PL license text and instead it says the project moved to GitHub and the MIT license is there. If MS-PL was what governed when the package was release then that seems right.

capfei:
Updated the PR and merging. Thank you!

PR #152, opened 2018-08-25
AE49_180825_105428.540

AE49:
Add MIT (https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/master/LICENSE)


capfei:
@AE49 I'm seeing the license as Apache-2.0 per https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/v2.23.302261847/LICENSE. Can you provide the me link to where it states MIT for this version? If it should Apache, let me know and I will update the PR.

ariel11:
@capfei - Is this curation for this package? https://www.nuget.org/packages/Microsoft.IdentityModel.Clients.ActiveDirectory/2.23.302261847?  If so, the download page does point to MIT.

fossygirl:
@capfei @ariel11 I've updated the YAML

PR #153, opened 2018-08-25
AE49_180825_122252.388

AE49:
Add MIT (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/m


capfei:
@AE49 For this version, I see the license as [GPL-2.0](https://github.com/icsharpcode/SharpZipLib/blob/0.86.0.518/doc/COPYING.txt). If you agree, I'll update PR.

ariel11:
@capfei - I agree it is Apache v2 for that version/date.

PR #156, opened 2018-08-25
AE49_180825_122738.226

AE49:
Add MIT (https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/master/LICENSE)


capfei:
The license file stated [MS-PL](https://github.com/System-IO-Abstractions/System.IO.Abstractions/tree/43e914bcf6f260d94955793e880eaceeba59416b) when this NuGet package version was published. If it states MIT somewhere else, please point me to it so I can take a look.

AE49:
The license link is broken on the NuGet page and the project site is under MIT which we discussed is not applicable after I submitted this entry. It can be declined.

PR #158, opened 2018-08-25
AE49_180825_144034.729

AE49:
Add Apache 2.0 for source code (https://github.com/dotnet/roslyn-analyzers/blob/master/License.txt)


capfei:
The NuGet package is under a Microsoft EULA. Closing PR.

PR #160, opened 2018-08-29
update all nuget curations to proper case

dabutvin:
* these curations happened before the latest crawler was running to respect case


dabutvin:
@capfei good catch!

I forgot we had open ones pending 🤷‍♂️ 
We will have to correct them all. I can take care of this though. 

Thanks

PR #128, opened 2018-08-25
AE49_180825_084134.800

AE49:
Add MIT (https://github.com/fsharp/FSharp.Compiler.Service/blob/master/LICENSE)

PR #163, opened 2018-08-31
AE49_180831_094647.464

AE49:
Add MIT (https://github.com/fsharp/FSharp.Compiler.Service/blob/master/LICENSE)


capfei:
Duplicate of #128

PR #306, opened 2018-09-14
AE49_180914_145915.598

AE49:
+ MIT https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/mas


capfei:
Duplicate of #305

PR #364, opened 2018-09-14
AE49_180914_203650.555

AE49:
Add MIT (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/m


capfei:
Duplicate of #303

PR #574, opened 2018-10-26
ariel11_181026_194736.657

ariel11:
**Type:** Missing

**Summary:**
Missing License

**Details:**
https://www.nuget.org/packages/Microsoft.PowerBI.Api/3.0.0 has no “License” or “Project” and is unlisted. 

**Resolution:**
The closest version is “MIT” - https://www.nuget.org/packages/Microsoft.PowerBI.Api/2.0.14. Declaring "MIT" but maybe we should leave blank?

**Affected definitions**:
- Microsoft.PowerBI.Api 3.0.0

### DIFF
--- a/curations/nuget/nuget/-/Microsoft.PowerBI.Api.yaml
+++ b/curations/nuget/nuget/-/Microsoft.PowerBI.Api.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: Microsoft.PowerBI.Api
+  provider: nuget
+  type: nuget
+revisions:
+  3.0.0:
+    licensed:
+      declared: MIT


fossygirl:
@capfei @ariel11 Do we need to talk about this further? Looks like it might be out of the bounds of our guidelines so I'd like to update them if possible. Let me know what we should do to resolve.

capfei:
I'm not sure with this one. The guidelines say to go with the nearest version, however, my holdup is that this version is unlisted.

fossygirl:
@jeffmcaffer Thoughts on what a good guideline in this case would be?

jeffmcaffer:
The reference version is listed and has a version so seems reasonable to use. Note that we can also look at the source code if it can be identified.

ariel11:
@jeffmcaffer - The version we are declaring the license for is "unlisted" - https://www.nuget.org/packages/Microsoft.PowerBI.Api/3.0.0.  Should we do anything special or different for "unlisted" packages?  There is no project link on the page to look at source code.

jeffmcaffer:
don't see any reason to treat unlisted packages differently. As long as the package can be accessed and used by teams then we have to deal with the licensing etc.

ariel11:
@capfei - I think Jeff answered the open question, above, yes?  Sounds like OK to check in MIT.  Let me know if I have any further action item here.  Thanks!

PR #620, opened 2018-11-06
AE49_181106_213349.529

AE49:
**Type:** Missing

**Summary:**
License

**Details:**
Add MIT

**Resolution:**
https://jquery.org/license/

**Affected definitions**:
- jQuery 2.1.3


capfei:
This is under both since this package contain both jQuery files, that are licensed under MIT, and custom code that is licensed under BSD-3-Clause per https://nugetpackages.codeplex.com.

PR #622, opened 2018-11-06
AE49_181106_213537.875

AE49:
**Type:** Missing

**Summary:**
License

**Details:**
Add MIT

**Resolution:**
https://github.com/Raynos/min-document/blob/v2.19.0/LICENCE

**Affected definitions**:
- min-document 2.19.0


AE49:
This is a duplicate.

PR #849, opened 2018-12-05
TravisEz13_181205_023338.509

TravisEz13:
**Type:** Missing

**Summary:**
Update the license for Microsoft.PowerShell.SDK

**Details:**
Update the license for Microsoft.PowerShell.SDK

**Resolution:**
https://github.com/PowerShell/PowerShell/blob/v6.1.0/LICENSE.txt

**Affected definitions**:
- Microsoft.PowerShell.SDK 6.1.0

### DIFF
--- a/curations/nuget/nuget/-/Microsoft.PowerShell.SDK.yaml
+++ b/curations/nuget/nuget/-/Microsoft.PowerShell.SDK.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: Microsoft.PowerShell.SDK
+  provider: nuget
+  type: nuget
+revisions:
+  6.1.0:
+    licensed:
+      declared: MIT


TravisEz13:
@fossygirl 	@capfei 	Is there a way to include this information in the package so that ClearlyDefined can automatically pick it up?  My team own this package.

fossygirl:
@TravisEz13 

The way our tooling currently picks up licenses in these cases is from a license file in the root of your nupkg. We haven't found one there right now, which is why the curation is needed. This would be the current remedy.

NuGet just published a new spec a couple weeks ago around license practices, and we haven't had a chance to implement that guidance yet. Once we do it may improve our tooling in specific cases like this enough that this won't happen again. 

@dabutvin for visibility or if he wants to add anything.

adityapatwardhan:
@fossygirl Is it possible for the tool to pick the license from the "LicenseUrl" field from the metadata.

fossygirl:
@adityapatwardhan @dabutvin or @jeffmcaffer will need to comment here on the current code functionality.

dabutvin:
@adityapatwardhan we do pick the licenseUrl from nuget to try and consider if we can determine the license automatically.

For this particular package it is set to `https://github.com/PowerShell/PowerShell/blob/master/LICENSE.txt`
We don't currently support this format of urls for determining the license.

We have recently considered fetching the raw data from this pattern of urls and running license detection on it to make a determination, but we don't do this today.
There is also some concern around pointing to the default branch of a repo and applying the license found there at the time of crawling for any version of the package as it can change over time.

The licenseUrl detection we support today can be seen [here](https://github.com/clearlydefined/service/blob/master/lib/utils.js#L170-L187)
For example if your package had a licenseUrl set to `https://opensource.org/licenses/MIT` we would pick it up as MIT.

dabutvin:
If you do decide to package your license with your nupkg for us to detect, you will need to make sure it is in a clean state.

Right now the license is not detectable - we use the same tooling as github to detect the license
You can tell because github UI shows 'View license' instead of 'MIT'

image


The [Newtonsoft.Json repo](https://github.com/jamesnk/newtonsoft.json) is a good example of a detectable MIT license
image

You can also see one of the new nuget concepts in use over [there](https://github.com/JamesNK/Newtonsoft.Json/blob/88dae521655b05bbe5b3d1821d82523a60d23dcc/Src/Newtonsoft.Json/Newtonsoft.Json.csproj#L23) `PackageLicenseExpression`
We don't support `PackageLicenseExpression` yet, but we will at some point

more info about the new nuget license handling coming: https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg-(Technical-Spec)

jeffmcaffer:
Not sure if the guidance from the NuGet team has been disseminated yet but as a blanked statement, all packages (nuget, npm, ...) *should* include the license text in the appropriate spot in the package (typically the root but YMMV). Since most licenses read something like "you must include a copy of this license with the software", if the license is not in the package, it is hard for the consumer to comply with the license. Putting it directly in the package is the golden path. 

Without that, ClearlyDefined would get the package, try to figure out the source location (ideally in GitHub) then go to GitHub, analyze the source for the corresponding commit and discover the license. We do that work now but there are weak links. It is often very hard to find the source for example.

jeffmcaffer:
As a meta point, thanks a TON for looking at this can taking action. Awesome.

PR #923, opened 2018-12-17
Update curations/git/github/gradle/gradle.yaml

capfei:


dabutvin:
Just a heads up, this one is not going to show up.
This is one of the components suffering from size too large issues
see https://github.com/clearlydefined/service/issues/254

PR #1107, opened 2019-03-04
Incorrect license for React files

bduranc:
**Type:** Incorrect

**Summary:**
Incorrect license for React files

**Details:**
License for <root>/src/main/resources/webapp/vendor/react/* was set incorrectly.
It should be BSD-2-Clause plus the explicit PATENTS grant.

**Resolution:**
PLEASE NOTE: SPDX does not actually appear to have an actual license identifier for the Facebook "BSD + PATENTS" licensing scenario. SPDX "BSD-2-Clause-Patents" is not the Facebook Grant of Additional Patent Rights Version 2 seen in older versions of React. But there is no other option to choose from.

**Affected definitions**:
- presto ba3080f3f7da6e7174f5d18d8b8ad5cb2e75c8a0

### DIFF
--- a/curations/git/github/prestodb/presto.yaml
+++ b/curations/git/github/prestodb/presto.yaml
@@ -1,0 +1,50 @@
+coordinates:
+  name: presto
+  namespace: prestodb
+  provider: github
+  type: git
+revisions:
+  ba3080f3f7da6e7174f5d18d8b8ad5cb2e75c8a0:
+    files:
+      - attributions:
+          - 'Copyright 2013-present, Facebook, Inc.'
+          - 'Copyright 2013-present, Facebook, Inc. All rights reserved.'
+        license: BSD-2-Clause-Patent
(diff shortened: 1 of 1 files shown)


capfei:
@bduranc 	If the license isn't in the SPDX list, then shouldn't it be "NOASSERTION" per the [curation guidelines](https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md)?

bduranc:
@capfei: Good point. I will make those changes.

capfei:
@bduranc, are you going to edit this PR, submit a new one, or I can update it for you? Let me know.

bduranc:
@capfei : It appears that the license for the React components are already updated, citing BSD 3 Clause+ "NOASSERTION".

We can close this PR as it appears the discrepancy has already been resolved.

image

Thanks.

capfei:
Thanks! 😃

PR #1109, opened 2019-03-07
Add JSON to license of twitter4j-core

richardfontana:
**Type:** Incorrect

**Summary:**
Add JSON to license of twitter4j-core

**Details:**
twitter4j-core includes some code under the JSON License. This is acknowledged in the LICENSE.txt file, for example; the full text of the license appears in a few source files. It is relevant that the JSON is widely considered to be a non-open-source license (and thus for example not somehow subsumable into the Apache License 2.0 for purposes of making a reasonably useful license description).

**Resolution:**
Correct the Declared license to "Apache-2.0 AND JSON"

**Affected definitions**:
- twitter4j-core 4.0.6

### DIFF
--- a/curations/maven/mavencentral/org.twitter4j/twitter4j-core.yaml
+++ b/curations/maven/mavencentral/org.twitter4j/twitter4j-core.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: twitter4j-core
+  namespace: org.twitter4j
+  provider: mavencentral
+  type: maven
+revisions:
+  4.0.6:
+    licensed:
+      declared: Apache-2.0 AND JSON


jeffmcaffer:
Great catch. One thing to note here is that this component was harvested some time ago before all the tools were in place. As a result, the tool-base information is lacking. As I type we are in the process of reprocessing all previously harvested components (including this one). 

As such, I'd like to defer merging this until after the component is crawled so we can see if the curation is actually needed. (I'll actually look to queue it earlier so we can review this sooner)

richardfontana:
I'm thinking about this some more after looking at https://github.com/clearlydefined/clearlydefined/blob/master/docs/clearly.md#licensed

It's good that ClearlyDefined is attempting to define what a "declared" license means and probably this could be improved and made more detailed. In this case, if the project maintainer had sort of "covered up" the presence of JSON, perhaps "Apache-2.0" would be appropriate for the declared license, but in this version the LICENSE.txt and readme.txt file go out of their way to mention that there's JSON-licensed subcomponents. The pom.xml file license element only mentions Apache-2.0. Depending on how "declared license" ought to be understood, you could make the case that JSON is not part of the declared license. Since ClearlyDefined will presumably disclose the presence of JSON-licensed source files anyway, it may not be problematic from an accuracy perspective to describe the declared license as Apache-2.0.

jeffmcaffer:
Thanks for the additional insightful comments/questions @richardfontana. We have tried to make the `declared` license be that which a normal consumer would see as the "envelope license". For example, the value in the `license` property of an npm or the license seen in the LICENSE file at the root of a repo. Most users would look at these as "this project is declaring that they are MIT".

The discovered licenses are those which are found/referenced in the random files in the component. The scoring model takes away points if your declared and discovered licenses don't match (at least in the `core` facet) so projects are encouraged to be inclusive/complete in their declarations.

We did have a discussion about packages that have both an explicit license metadata attribute of some sort and a license file. Which should win. I'd love to hear the opinion of others (and frankly I can't remember what we decided) but it has to go one way or the other.

richardfontana:
@jeffmcaffer I am very interested in this concept of the declared license (and the distinction from the discovered license(s)) and it could be something I could be motivated to help with in the context of this project. In brief I think it may be more of a can of worms than you may be assuming. Which is not to say it is not valid as something to try to identify. But I should probably comment further outside of this specific conversation.  

In this case, a normal consumer might see the license as Apache-2.0 because they might be inclined to ignore the "fine print" that references the presence of the JSON license and they might be inclined not to do a license scan (or human eyeball equivalent) on all the source code. That was probably the intention of the maintainer; they probably struggled over this issue of whether to sweep the problem under the rug or relegate it to "fine print". (Recent versions of Twitter4J have removed the JSON dependencies. There could be an argument that resources like ClearlyDefined can help motivate maintainers to fix problematic things like this.) 

A more careful consumer however would notice that fine print. After all it's right there at the bottom of the LICENSE file, but in a way that no license scanner is likely to be configured to catch. A still more careful consumer would find JSON licensed source files anyway. My main point is just that it's not obvious what the declared license is, because "declared license" is a made-up concept, not having any legal validity, that reflects a real psychological phenomenon in FOSS, a desire to have relatively abstract ways of describing licenses of packages. Anyway I will try to provide further thoughts on this elsewhere.

jeffmcaffer:
For better or for worse, other systems like FOSSology have more or less the equivalent concept (https://www.fossology.org/get-started/basic-workflow/#bw5). I agree that some consumers may confuse the Declared license in the same way as they would confuse "envelope" licenses today. ClearlyDefined attempts to highlight that condition subtracting points from the score if the declared licenses do not match the discovered licenses. This is a signal that something is amiss and further investigation is required. 

Of course, consumers are free to simply use the discovered license set (it is a list of expressions) as the effective license and ignore the declared license altogether. 

In any event, your feedback is great and very much appreciated. Keep it coming in whatever forum you choose.

As it is for this PR, my sense is that your correction stands and is the valid as the license file in the component does mention the JSON license and that was simply not picked up by the tools. Correct?

fossygirl:
@richardfontana @jeffmcaffer What would you folks like to do with this curation at this point?

fossygirl:
@capfei Can you review this curation for accuracy, and merge if appropriate?

PR #1116, opened 2019-03-11
Declared license should be BSD-3-Clause

richardfontana:
**Type:** Incorrect

**Summary:**
Declared license should be BSD-3-Clause

**Details:**
Text of root directory license closely matches SPDX BSD-3-Clause (but possibly not a match in an SPDX-conformance sense because of an overly strict SPDX definition); in any case it is not BSD-2-Clause. docs/license.rst also says it is licensed under "a 3-clause BSD license"

**Resolution:**
Correct declared license to BSD-3-Clause

**Affected definitions**:
- Flask 0.10.1

### DIFF
--- a/curations/pypi/pypi/-/Flask.yaml
+++ b/curations/pypi/pypi/-/Flask.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: Flask
+  provider: pypi
+  type: pypi
+revisions:
+  0.10.1:
+    licensed:
+      declared: BSD-3-Clause


dabutvin:
nice catch. I found where this was coming from and opened an issue on the crawler repo to address it
https://github.com/clearlydefined/crawler/issues/273

capfei:
Let me know if I can process this PR or leave it open until the issue is resolved.

dabutvin:
@capfei  you can process this PR. No need to wait

PR #1125, opened 2019-03-12
License

AE49:
**Type:** Missing

**Summary:**
License

**Details:**
Add BSD-2-Clause

**Resolution:**
License noted on Maven download page

**Affected definitions**:
- protobuf-java 3.0.0

### DIFF
--- a/curations/maven/mavencentral/com.google.protobuf/protobuf-java.yaml
+++ b/curations/maven/mavencentral/com.google.protobuf/protobuf-java.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: protobuf-java
+  namespace: com.google.protobuf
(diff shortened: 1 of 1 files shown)


capfei:
@AE49 	, I would say the license is BSD-3 instead of BSD-2. When I look in the JAR, the license in the file headers are BSD-3 and the source repo has BSD-3.

https://github.com/protocolbuffers/protobuf/tree/v3.0.0
https://search.maven.org/remotecontent?filepath=com/google/protobuf/protobuf-java/3.0.0/protobuf-java-3.0.0.jar

Let me know if you are agree that this should be updated.

AE49:
@capfei I agree with your findings. Thanks.

PR #1218, opened 2019-04-10
Apache

carlbrochu:
**Type:** Missing

**Summary:**
Apache

**Details:**
https://www.nuget.org/packages/CsvHelper has dual licenses

**Resolution:**
https://www.nuget.org/packages/CsvHelper

**Affected definitions**:
- CsvHelper 12.1.2


royaljust:
@capfei - since this component is "dual licensed" don't we curate as "OR" rather than "AND"?

PR #1260, opened 2019-04-25
Removing licenses other than BSD 3 from LICENSE

fossygirl:
**Type:** Ambiguous

**Summary:**
Removing licenses other than BSD 3 from LICENSE

**Details:**
Component says it has multiple licenses from the license file, but I believe it's getting that data from other components tough-cookie depends on. 

**Resolution:**
I believe this is licensed only under BSD-3 per https://github.com/salesforce/tough-cookie/blob/74e59de50b719bb9a1b01c8c9db57fa31401ed1a/LICENSE

**Affected definitions**:
- tough-cookie 1.2.0

### DIFF
--- a/curations/npm/npmjs/-/tough-cookie.yaml
+++ b/curations/npm/npmjs/-/tough-cookie.yaml
@@ -8,5 +8,5 @@ revisions:
       - attributions:
           - Copyright (c) 2007
           - 'Copyright (c) 2015, Salesforce.com, Inc.'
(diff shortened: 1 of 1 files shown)


capfei:
@fossygirl the files that state they use the different licenses are included in the NPM package. Wouldn't that mean that it should be listed as BSD-3 AND MPL-1.1 OR GPL-2.0-or-later OR LGPL-2.1-or-later AND CC0-1.0?

fossygirl:
Thanks @capfei. I updated the YAML but I'm not 100% sure I did it right. Could you verify?

PR #1778, opened 2019-05-13
License

AE49:
**Type:** Incomplete

**Summary:**
License

**Details:**
Add MIT

**Resolution:**
https://github.com/jamiebuilds/create-react-context/blob/v0.2.3/LICENSE

**Affected definitions**:
- create-react-context 0.2.3

### DIFF
--- a/curations/npm/npmjs/-/create-react-context.yaml
+++ b/curations/npm/npmjs/-/create-react-context.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   0.2.3:
     licensed:
-      declared: NOASSERTION
+      declared: MIT


capfei:
@AE49 Not sure this should be MIT since it is modified and adds the restrictions on who can use it. I would leave it as "No Assertion", but would love to hear others input.

AE49:
I've been informed by Cliff Allen the restriction in the Jamie Builds was lifted.

capfei:
The license was updated two weeks after this NPM version was published so I wouldn't treat this curation any differently. Since the license in the NPM package and the tagged source version contain the restricted license, it should be NO ASSERTION.

ariel11:
@capfei - I agree that leaving this version as NOASSERTION makes sense.

capfei:
For now, I am closing this PR.

PR #1858, opened 2019-05-31
Update project and file level licenses

bduranc:
**Type:** Incomplete

**Summary:**
Update project and file level licenses

**Details:**
The top-level license for this component should be 'LGPL-3.0-or-later'.

Some files were mistakenly identified as 'LGPL-2.0-or-later' when the license headers say '3.0-or-later'. Files following the pattern ''psycopg/adapter*.{h,c}'' and ''psycopg/microprotocol*.{h,c}'' are dual-licensed 'LGPL OR Zlib'.

**Resolution:**
Updated project level license as well as individual file licenses where incorrect discovery occurred. Files which are dual-licensed under Zlib have also been updated to reflect this... 

https://github.com/psycopg/psycopg2/blob/2_8_2/LICENSE (See "Alternative licenses" at bottom) 

**Affected definitions**:
- psycopg2 2.8.2

### DIFF
--- a/curations/pypi/pypi/-/psycopg2.yaml
+++ b/curations/pypi/pypi/-/psycopg2.yaml
@@ -12,3 +12,150 @@ revisions:
         revision: 324cded1661995cc006e04e2c5e6b00e61ab74ba
         type: git
         url: 'https://github.com/psycopg/psycopg2/commit/324cded1661995cc006e04e2c5e6b00e61ab74ba'
+    files:
+      - attributions:
+          - Copyright (c) 2003-2019 Federico Di Gregorio <fog@debian.org>
+        license: LGPL-3.0-or-later
+        path: psycopg2-2.8.2/setup.py
+      - attributions:
+          - 'Copyright (c) 2007 Free Software Foundation, Inc. <http://fsf.org/>'
+        license: LGPL-3.0
+        path: psycopg2-2.8.2/doc/COPYING.LESSER
+      - attributions:
+          - Copyright (c) 2012-2019 Daniele Varrazzo <daniele.varrazzo@gmail.com>
+        license: LGPL-3.0-or-later
+        path: psycopg2-2.8.2/lib/_json.py
+      - attributions:
+          - Copyright (c) 2003-2019 Federico Di Gregorio <fog@debian.org>
+        license: LGPL-3.0-or-later OR Zlib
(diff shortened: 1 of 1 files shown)


fossygirl:
Hi folks - I'd like to see how review of this is going. Do we have any questions that still need answering here? If so, @bduranc could you work with @ariel11 on this? Don't want this to linger without either some issues identified, merge, or close by the 30 days mark.

ariel11:
@fossygirl - No questions at the moment.  I grabbed it from the end of the queue because of the 30 day goal.  However, I have to double-check each file and you can see the diff is 161 lines. Im working on it.  Thanks

fossygirl:
Great, thanks for the update, @ariel11!

ariel11:
@bduranc - noting that https://pypi.org/project/psycopg2/2.8.2/ says the “license” is “GNU Library or Lesser General Public License (LGPL), Zope Public License (LGPL with exceptions or ZPL).”  However, I don’t see any reference to the Zope Public License in the source repo.

As you mention above, if you go to the source for tagged version 2.8.2, the license is https://raw.githubusercontent.com/psycopg/psycopg2/324cded1661995cc006e04e2c5e6b00e61ab74ba/LICENSE, which says “LGPLv3 or later” with a “special exception” for linking to OpenSSL  OR a bsd-like license, which is the Zlib license. 

It appears that maybe the reference to the ZPL is a mistake on https://pypi.org/project/psycopg2/2.8.2/ ...?

bduranc:
@ariel11: That is my theory as well.

If you look at the history of LICENSE.MD, you will see that older versions as far back as 2.4.0 had some mention of the ZPL as an alternative license (see: https://github.com/psycopg/psycopg2/commit/611606d53294ec68ab61aea9ba9149d3c6da4750#diff-9879d6db96fd29134fc802214163b95a). 

I cannot find any obvious mention of the same in 2.8.2.

ariel11:
@bduranc - what do you make of the mention of the Zope license on this file - https://github.com/psycopg/psycopg2/blob/324cded1661995cc006e04e2c5e6b00e61ab74ba/setup.py (row 59)?  It's the only reference I saw on the files curated in this PR.

ariel11:
@bduranc - I am going to change the "Declared" license to "OTHER" because the license for this version says LGPL with a special exception and there is not a SPDX for that scenario. I merged your other file level curations.  Thanks

PR #1863, opened 2019-06-04
Update file-level discovered licenses

bduranc:
**Type:** Incorrect

**Summary:**
Update file-level discovered licenses

**Details:**
Some of the discovered licenses at the file-level are incorrect.

**Resolution:**
Update cases of incorrect file-level licenses. Here, these are files that are discovered NOASSERTION or multi-license (AND) when they should actually be dual-license (OR).

(See bottom of https://github.com/stuk/jszip/)

**Affected definitions**:
- jszip 265c959b1f15f8cd5ebe152154b88796e4a82f42

### DIFF
--- a/curations/git/github/stuk/jszip.yaml
+++ b/curations/git/github/stuk/jszip.yaml
@@ -29,7 +29,7 @@ revisions:
       - license: MIT OR GPL-3.0
         path: bower.json
       - attributions:
-          - Copyright (c) 1989 - 2007 PKWARE Inc.
+          - Copyright (c) 1989 - 2007 PKWARE Inc., All Rights Reserved.
         license: OTHER
         path: docs/APPNOTE.TXT
       - attributions:
--- a/curations/git/github/stuk/jszip.yaml
+++ b/curations/git/github/stuk/jszip.yaml
@@ -1,0 +1,45 @@
+coordinates:
+  name: jszip
+  namespace: stuk
+  provider: github
+  type: git
+revisions:
+  265c959b1f15f8cd5ebe152154b88796e4a82f42:
+    files:
+      - license: MIT OR GPL-3.0
+        path: index.html
+      - attributions:
+          - (c) 1995-2013 Jean-loup Gailly and Mark Adler
+          - (c) 2014-2017 Vitaly Puzrin and Andrey Tupitsin
+          - (c) 2009-2016 Stuart Knightley <stuart at stuartk.com>
+        license: MIT OR GPL-3.0
+        path: dist/jszip.js
+      - license: MIT OR GPL-3.0
+        path: README.markdown
+      - license: MIT
+        path: vendor/FileSaver.js
+      - license: MIT OR GPL-3.0
+        path: component.json
+      - license: OTHER
+        path: lib/zipEntry.js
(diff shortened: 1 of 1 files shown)


bduranc:
The file **<root>/lib/zipEntry.js** was discovered as "NOASSERTION" initially by the tooling, and set as "OTHER" by me. But upon further inspection there are no license details in the file, so value should be "NONE". I can create a new curation for this one change if required.

ariel11:
@bduranc - do you think the curation for this file - https://github.com/Stuk/jszip/blob/265c959b1f15f8cd5ebe152154b88796e4a82f42/dist/jszip.js - would be (MIT OR GPLv3) per row 7 "**AND Zlib**" per row 5687-5701 (and in other sections of this file)?

ariel11:
@bduranc, @royaljust, or @AE49  - two other files I have questions on.

1. https://github.com/Stuk/jszip/blob/265c959b1f15f8cd5ebe152154b88796e4a82f42/test/jquery-1.8.3.min.js - this is curated as MIT, but the file points to "jquery.org/license," which says multiple things, so do you think this would be "OTHER?"
2. https://github.com/Stuk/jszip/blob/265c959b1f15f8cd5ebe152154b88796e4a82f42/lib/license_header.js - this is curated as "MIT OR GPL-3.0," but since there are two different things referenced on this file, do you think it would be (MIT OR GPL-3.0) **AND MIT**?

bduranc:
> @bduranc - do you think the curation for this file - https://github.com/Stuk/jszip/blob/265c959b1f15f8cd5ebe152154b88796e4a82f42/dist/jszip.js - would be (MIT OR GPL-3.0) per row 7 "**AND Zlib**" per row 5687-5701 (and in other sections of this file)?

BD: @ariel11 : Yes, the license should also include ZLib. 

> @bduranc, @royaljust, or @AE49 - two other files I have questions on.
> 
> 1. https://github.com/Stuk/jszip/blob/265c959b1f15f8cd5ebe152154b88796e4a82f42/test/jquery-1.8.3.min.js - this is curated as MIT, but the file points to "jquery.org/license," which says multiple things, so do you think this would be "OTHER?"
> 2. https://github.com/Stuk/jszip/blob/265c959b1f15f8cd5ebe152154b88796e4a82f42/lib/license_header.js - this is curated as "MIT OR GPL-3.0," but since there are two different things referenced on this file, do you think it would be (MIT OR GPL-3.0) **AND MIT**?

BD: @ariel11 : For the first one, looks to me like the core of jQuery is licensed under MIT. But sample code is under CC0. I don't believe the Sample Code portion affects the compressed file under review, but perhaps someone else knows different.

For the second point, that is a good question. I'd be interested in hearing other people's thoughts. 
I can see why adding **AND MIT** would make sense as Pako is a separate component.

I think this shows that we need to improve the visibility between the license expression and the actual line(s) of code that were identified. In the case of the first file, there is some Zlib-licensed code but it is lower down in the file and not in the header.

ariel11:
@bduranc - We just had a steering committee call and agreed to curate "(MIT OR GPL-3.0) AND MIT," as discussed above.  I will update the curation and merge.  Thanks!!

PR #1864, opened 2019-06-04
Update file-level discovered licenses

bduranc:
**Type:** Incorrect

**Summary:**
Update file-level discovered licenses

**Details:**
Some of the discovered licenses at the file-level are incorrect.

**Resolution:**
Update cases of incorrect file-level licenses. Here, these are files that are discovered NOASSERTION or multi-license (AND) when they should actually be dual-license (OR).

(See bottom of https://github.com/stuk/jszip/)

**Affected definitions**:
- jszip 3.2.0

### DIFF
--- a/curations/npm/npmjs/-/jszip.yaml
+++ b/curations/npm/npmjs/-/jszip.yaml
@@ -1,0 +1,24 @@
+coordinates:
+  name: jszip
+  provider: npmjs
+  type: npm
+revisions:
+  3.2.0:
+    files:
+      - license: MIT
+        path: package/vendor/FileSaver.js
+      - attributions:
+          - (c) 2009-2016 Stuart Knightley <stuart at stuartk.com>
+        license: MIT OR GPL-3.0
+        path: package/lib/license_header.js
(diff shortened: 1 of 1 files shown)


bduranc:
The file **<root>/package/lib/zipEntry.js** was discovered as "NOASSERTION" initially by the tooling, and set as "OTHER" by me. But upon further inspection there are no license details in the file, so value should be "NONE". I can create a new curation for this one change if required.

ariel11:
@bduranc - I have a question on curating the copyright that is attached to the license vs. the code?  For example, on the [LICENSE.markdown file](https://clearlydefined.io/file/14450c78405ad2a2173e25740b56406556779149df9c4c83523a8c63d0686210), curating the copyright from the GPLv3 license itself (https://www.gnu.org/licenses/gpl-3.0.txt).  @royaljust - do you have any thoughts here?

ariel11:
@bduranc - I am removing the copyright that applies to the GPLv3 license template and merging this PR.  Thanks!

PR #1888, opened 2019-06-07
Declared License

AE49:
**Type:** Incomplete

**Summary:**
Declared License

**Details:**
Dual licensed as noted on Npm page and GitHub repo so I have added Apache 2.0 or MIT, although I only found the Apache 2.0 license file in the repo for this version.

**Resolution:**
https://git.coolaj86.com/coolaj86/btoa.js/src/tag/v1.1.2/LICENSE

**Affected definitions**:
- [btoa 1.1.2](https://clearlydefined.io/definitions/npm/npmjs/-/btoa/1.1.2/1.1.2)

### DIFF
--- a/curations/npm/npmjs/-/btoa.yaml
+++ b/curations/npm/npmjs/-/btoa.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   1.1.2:
     licensed:
-      declared: Apache-2.0 OR MIT
+      declared: Apache-2.0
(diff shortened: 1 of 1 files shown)


ariel11:
@AE49 - Per our call, I think we decided there is more evidence the license is only Apache given the package.json (https://clearlydefined.io/file/e01de66e7bff743dc240c05ee06940a0487d22af5f01cbc680a2becc3193ea94) and there only being an Apache license file - correct?  If we learn later that it should be "or MIT" we can curate that.

PR #1921, opened 2019-06-08
Declared License

AE49:
**Type:** Incomplete

**Summary:**
Declared License

**Details:**
Public Domain

**Resolution:**
https://github.com/dchest/tweetnacl-js/blob/v0.14.1/COPYING.txt

**Affected definitions**:
- [tweetnacl 0.14.1](https://clearlydefined.io/definitions/npm/npmjs/-/tweetnacl/0.14.1/0.14.1)

### DIFF
--- a/curations/npm/npmjs/-/tweetnacl.yaml
+++ b/curations/npm/npmjs/-/tweetnacl.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: tweetnacl
+  provider: npmjs
+  type: npm
+revisions:
+  0.14.1:
+    licensed:
+      declared: NONE


ariel11:
@AE49 - Changing to "Other" for public domain/CC-0.

ariel11:
@AE49 - This one is unclear to me. The NPM has "Unlicense" in the "license" field (https://www.npmjs.com/package/tweetnacl/v/0.14.1). While the "Copying" file looks like something else (https://github.com/dchest/tweetnacl-js/blob/v0.14.1/COPYING.txt) - not sure if this is a SPDX license. And then it looks like the COPYING.txt file is curated as CC-0 in ClearlyDefined (https://clearlydefined.io/definitions/npm/npmjs/-/tweetnacl/0.14.1/0.14.1).

pombredanne:
The license in https://github.com/dchest/tweetnacl-js/blob/v0.14.1/package.json is "SEE LICENSE in COPYING.txt" and https://github.com/dchest/tweetnacl-js/blob/v0.14.1/COPYING.txt  is a clear public domain dedication BUT NOT a CC0-1.0 (it is  somewhat close to a notice for the CC0-1.0 but falls short of actually naming the license) ...whatever license you would see on https://www.npmjs.com/package/tweetnacl/v/0.14.1 would be for the latest version and the same applies to https://github.com/dchest/tweetnacl-js

So both Licensee as seen at https://github.com/dchest/tweetnacl-js  the NPM registry web pages are wrong. (or rather only apply to whatever the current latest version is)

This version is dedicated to the public domain... which once again is not known to SPDX, hence another case for OTHER :|

These "OTHER" are really annoying ... they are a waste of curation energy. As we are losing the clear fact this is dedicated to the public domain :|

pombredanne:
I rebased this and used OTHER for the license based on the rationale explained in https://github.com/clearlydefined/curated-data/pull/1921/#issuecomment-508889325

PR #1924, opened 2019-06-08
Declared License

AE49:
**Type:** Incomplete

**Summary:**
Declared License

**Details:**
Add MIT or any version of LGPL

**Resolution:**
https://github.com/jindw/xmldom/blob/master/LICENSE

**Affected definitions**:
- [xmldom 0.1.22](https://clearlydefined.io/definitions/npm/npmjs/-/xmldom/0.1.22/0.1.22)

### DIFF
--- a/curations/npm/npmjs/-/xmldom.yaml
+++ b/curations/npm/npmjs/-/xmldom.yaml
@@ -3,6 +3,9 @@ coordinates:
   provider: npmjs
   type: npm
 revisions:
+  0.1.22:
+    licensed:
+      declared: MIT OR LGPL-2.0-or-later
   0.1.27:
     licensed:
       declared: LGPL-2.0 OR MIT


ariel11:
@AE49 - I am not sure what to do when the license field says "none" and then the package is an untagged version in the repo.  Also, the link http://www.gnu.org/licenses/lgpl.html resolves to LGPL-3.0, so we should confirm if that should be the license or not.

AE49:
@ariel11 - Ok, and if you want others to comment on the "None". Although there is None on the NuGet page, there is a direct link to the GitHub repo with MIT and LGPL.

The LGPL version isn't specified in the GitHub repo (https://github.com/jindw/xmldom/blob/master/LICENSE) and for this reason I looked at the LGPL license and applied this - 

If the Library as you received it specifies that a certain numbered version of the GNU Lesser General Public License “or any later version” applies to it, you have the option of following the terms and conditions either of that published version or of any later version published by the Free Software Foundation. If the Library as you received it does not specify a version number of the GNU Lesser General Public License, you may choose any version of the GNU Lesser General Public License ever published by the Free Software Foundation.

I can also see why you would say it's LGPL v3 from the URL - https://www.gnu.org/licenses/lgpl-3.0.html.

PR #2131, opened 2019-06-27
Missing Declared license

ariel11:
**Type:** Incomplete

**Summary:**
Missing Declared license

**Details:**
This one has some inconsistencies. 4 licenses are listed here: https://mvnrepository.com/artifact/com.github.jnr/jnr-posix/3.0.45. Note it says EPL 1.0.

**Resolution:**
However, on the code header it says "The contents of this file are subject to the Eclipse Public
License Version 2.0 (the "License")" but then the link it gives for the license goes to the CPL 1.0.  So, I didn't curate EPL 1.0 and wasn't sure to curate CPL 1.0 AND EPL 2.0? 

**Affected definitions**:
- [jnr-posix 3.0.45](https://clearlydefined.io/definitions/maven/mavencentral/com.github.jnr/jnr-posix/3.0.45/3.0.45)

### DIFF
--- a/curations/maven/mavencentral/com.github.jnr/jnr-posix.yaml
+++ b/curations/maven/mavencentral/com.github.jnr/jnr-posix.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   3.0.45:
     licensed:
-      declared: CPL-1.0 AND EPL-2.0 OR (LGPL-2.1-or-later OR GPL-2.0-or-later)
+      declared: CPL-1.0 AND EPL-2.0 OR (LGPL-2.1 OR GPL-2.0)
--- a/curations/maven/mavencentral/com.github.jnr/jnr-posix.yaml
+++ b/curations/maven/mavencentral/com.github.jnr/jnr-posix.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: jnr-posix
+  namespace: com.github.jnr
+  provider: mavencentral
+  type: maven
+revisions:
+  3.0.45:
+    licensed:
+      declared: CPL-1.0 AND EPL-2.0 OR (LGPL-2.1-or-later OR GPL-2.0-or-later)


ariel11:
See file header:

/***** BEGIN LICENSE BLOCK *****
 * Version: EPL 2.0/GPL 2.0/LGPL 2.1
 *
 * The contents of this file are subject to the Eclipse Public
 * License Version 2.0 (the "License"); you may not use this file
 * except in compliance with the License. You may obtain a copy of
 * the License at http://www.eclipse.org/legal/cpl-v10.html
 *
 * Software distributed under the License is distributed on an "AS
 * IS" basis, WITHOUT WARRANTY OF ANY KIND, either express or
 * implied. See the License for the specific language governing
 * rights and limitations under the License.
 *
 * Alternatively, the contents of this file may be used under the terms of
 * either of the GNU General Public License Version 2 or later (the "GPL"),
 * or the GNU Lesser General Public License Version 2.1 or later (the "LGPL"),
 * in which case the provisions of the GPL or the LGPL are applicable instead
 * of those above. If you wish to allow use of your version of this file only
 * under the terms of either the GPL or the LGPL, and not to allow others to
 * use your version of this file under the terms of the CPL, indicate your
 * decision by deleting the provisions above and replace them with the notice
 * and other provisions required by the GPL or the LGPL. If you do not delete
 * the provisions above, a recipient may use your version of this file under
 * the terms of any one of the CPL, the GPL or the LGPL.
 ***** END LICENSE BLOCK *****/

AE49:
@fossygirl @bduranc @royaljust Please take a look at this curation and provide your input. Thank you.

bduranc:
I also encountered this same license when curating a version of JRuby: 

https://github.com/jruby/jruby/blob/8a269e34ff3d242ec90fb08061f79c2757a5e157/lib/ruby/stdlib/ffi/union.rb 

It seems to me like at some point they moved from the CPL over to the EPL as this was quite common, but forgot to change all the text.  

I'm interested also to see what the others think.

ariel11:
Thanks, @bduranc.  In this instance, I curated what I did from the view point that we should not be interpreting what someone meant to be the license.  However, I looked at two instances yesterday where there were inconsistencies between the package licensing metadata and the licensing info in the source.  That happened here as well, as I ignored the reference to EPL 1.0 in the package metadata (the license tags here: https://mvnrepository.com/artifact/com.github.jnr/jnr-posix/3.0.45), because EPL 1.0 wasn't mentioned in the source for this version. And then here as well, it does look like the mismatch between the license name given for EPL 2.0 and the license link (for CPL 1.0) was an oversight.  When curating,It would be good to decide what to do when we have evidence that supports a licensing 'error.'

royaljust:
Interesting issue.  I don't think we should be curating based on suspicion of error, even when the error seems pretty clear.  In this case, I would treat the declared license as what the Maven repo says and note the discovered EPL 2.0 in the source.  Any thoughts on this @pombredanne?

AE49:
@ariel11 Did we agree as a group to Declare EPL 1.0 and EPL 2.0 would be a discovered license?

ariel11:
@AE49 - My understanding from @royaljust is that because there is not a "license" file in the jar, we will carry through what the metadata says for the maven package.  And that we are not going to use the license block on some of the files to curate the 'declared' field.  So, my understanding is the curation for the 'declared' field becomes the (4) licenses tagged at the top here: https://mvnrepository.com/artifact/com.github.jnr/jnr-posix/3.0.45. An open question for me then is whether we carry through the "or later" for GPL/LGPL because that info was found on some of the files and not in a canonical location for the overall project.

AE49:
@ariel11 Yes, I remember we were going with the four licenses. My question is specific to the EPL license. We can discuss your last question with @royaljust as we need to make a decision.

PR #2162, opened 2019-07-01
Declared License

AE49:
**Type:** Incomplete

**Summary:**
Declared License

**Details:**
Add MIT

**Resolution:**
NuGet license is MIT, closest version 3.16.0 with MIT license (https://github.com/AzureAD/azure-activedirectory-library-for-dotnet)/blob/ver3.16.0/LICENSE

**Affected definitions**:
- [Microsoft.IdentityModel.Clients.ActiveDirectory 3.16.1](https://clearlydefined.io/definitions/nuget/nuget/-/Microsoft.IdentityModel.Clients.ActiveDirectory/3.16.1/3.16.1)

### DIFF
--- a/curations/nuget/nuget/-/Microsoft.IdentityModel.Clients.ActiveDirectory.yaml
+++ b/curations/nuget/nuget/-/Microsoft.IdentityModel.Clients.ActiveDirectory.yaml
@@ -30,6 +30,9 @@ revisions:
   3.14.2:
     licensed:
       declared: MIT
+  3.16.1:
+    licensed:
+      declared: MIT
   3.17.2:
     licensed:
       declared: MIT


ariel11:
@ignacionr - can you please assist with this failed check?

ignacionr:
This one is very interesting. Supposedly, with the changes I introduced with the tool, this should have been solved (on the underlying, starting point of this PR) but I don't see that it requires to update the baseline. So something must have happened. Let me check...

ignacionr:
Actually I can see that Gene already resolved this by modifying the definition manually, 27 days ago: https://github.com/clearlydefined/curated-data/commit/d234e131553d089b503d32f10c33798f25d79f87

AE49:
@ignacionr I don't see that Ariel has the ability to merge the PR.

ignacionr:
@AE49 that's the interesting thing; the changes on this PR are based on an older version of the file they modify, and GitHub normally would allow you to rebase changes (merge the new baseline), but I don't see that button being offered here. Let me show you so you can understand what's happening.

- this is the current master branch file, notice the line with a name for the package https://github.com/clearlydefined/curated-data/blob/master/curations/nuget/nuget/-/Microsoft.IdentityModel.Clients.ActiveDirectory.yaml#L21 
- this is the line that these changes are proposed upon https://github.com/clearlydefined/curated-data/pull/2162/files#diff-64f4ed4ba03b902d2d39f892212e11afR21 

The contents of the baseline files should either be the same, or Github should offer you to refresh by merging. But the former is not true, and the latter doesn't happen.

ignacionr:
OK so for all pending PRs, we need to rebase the changes (merge from master).

PR #2181, opened 2019-07-03
Updating curation

ariel11:
**Type:** Incorrect

**Summary:**
Updating curation

**Details:**
Per team call, we want to curate https://mvnrepository.com/artifact/com.github.jnr/jnr-posix/3.0.45 to match the license metadata on the mvnrepository. 

**Resolution:**
So that would be: CPL 1.0 OR EPL 1.0 OR GPL 2.0 OR LGPL 2.1.  The tooling won't let me enter it this way, so will need to be fixed on the .yaml.

**Affected definitions**:
- [jnr-posix 3.0.45](https://clearlydefined.io/definitions/maven/mavencentral/com.github.jnr/jnr-posix/3.0.45/3.0.45)

### DIFF
--- a/curations/maven/mavencentral/com.github.jnr/jnr-posix.yaml
+++ b/curations/maven/mavencentral/com.github.jnr/jnr-posix.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   3.0.45:
     licensed:
-      declared: (EPL-1.0 OR CPL-1.0) AND (GPL-2.0 OR LGPL-2.1)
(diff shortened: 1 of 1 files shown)


ariel11:
@AE49 - OK, submitted this new curation.  The tooling wouldn't let me put all "OR's," so I updated the .yaml to say "declared: EPL-1.0 OR CPL-1.0 OR GPL-2.0 OR LGPL-2.1," per our team meeting.  If the author clarifies the discrepancy between this licensing metadata and the license block on some of the files, we can update again.

PR #2192, opened 2019-07-08
Corrected license info for NetMQ.

cgranade:
**Type:** Incorrect

**Summary:**
Corrected license info for NetMQ.

**Details:**
The license was marked as NOTDECLARED.

**Resolution:**
The license was updated to the correct definition of LGPL 3.0 (exact).

**Affected definitions**:
- [NetMQ 4.0.0.1](https://clearlydefined.io/definitions/nuget/nuget/-/NetMQ/4.0.0.1)

### DIFF
--- a/curations/nuget/nuget/-/NetMQ.yaml
+++ b/curations/nuget/nuget/-/NetMQ.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   4.0.0.1:
     licensed:
-      declared: LGPL-3.0-only
+      declared: OTHER
--- a/curations/nuget/nuget/-/NetMQ.yaml
+++ b/curations/nuget/nuget/-/NetMQ.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: NetMQ
+  provider: nuget
+  type: nuget
+revisions:
+  4.0.0.1:
+    licensed:
+      declared: LGPL-3.0-only


pombredanne:
@cgranade thank you ++ for this
You may not have noticed but https://github.com/zeromq/netmq/blob/4.0.0.1/COPYING.LESSER#L171 has some extra terms at the end. 

The exception is this https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/zeromq-exception-lgpl-3.0.LICENSE but it is not known from SPDX.

Also it looks like all the notices in source code file point to an `LGPL-3.0-or-later` rather than a bare `LGPL-3.0-only`.

ariel11:
@pombredanne - The extra language is so close to https://spdx.org/licenses/Classpath-exception-2.0.html#licenseExceptionText.  Here's the delta.
image.  It seems 'wrong' to curate this as "OTHER" due to this extra custom text, but that's what we have been doing in other instances...  @royaljust, thoughts?

Also, the ReadMe just says "This is an open source project licensed under the LGPLv3," so I would not have set the declared license to "or later" because that is not stated for the project level (just on some files).

ariel11:
I would be included to curate this as "LGPL-3.0-only AND OTHER," where OTHER is for that extra text.  And I would do LGPL-3.0-only because the readme says "This is an open source project licensed under the LGPLv3."  Any files that say "or later' could be curated as such for that file.  Thoughts?

ariel11:
Changing Declared license to "OTHER" due to non-SPDX recognized language on the license file.

PR #2195, opened 2019-07-09
Declared License

AE49:
**Type:** Incomplete

**Summary:**
Declared License

**Details:**
Add MIT

**Resolution:**
https://github.com/nodejs/node/blob/7e463500b2483c084d1dbc016806d6f1092ddabf/LICENSE

**Affected definitions**:
- [node 7e463500b2483c084d1dbc016806d6f1092ddabf](https://clearlydefined.io/definitions/git/github/nodejs/node/7e463500b2483c084d1dbc016806d6f1092ddabf/7e463500b2483c084d1dbc016806d6f1092ddabf)

### DIFF
--- a/curations/git/github/nodejs/node.yaml
+++ b/curations/git/github/nodejs/node.yaml
@@ -16,6 +16,9 @@ revisions:
           - 'Copyright (c) 2007, 2008 gnombat@users.sourceforge.net'
         license: GPL-3.0
         path: doc/sh_main.js
+  7e463500b2483c084d1dbc016806d6f1092ddabf:
+    licensed:
+      declared: MIT
   b0e5f195dfce3e2b99f5091373d49f6616682596:
     files:
       - attributions:


pombredanne:
@AE49 Isn't there more than just MIT in node?

AE49:
@pombredanne I was adding the Declared License. I see other licenses for external maintained libraries - wouldn't these be Discovered Licenses?

pombredanne:
@AE49 good point but IMHO the declared license is not the "primary" license.

I would tend to think that when we curate all licenses should be reported there in all cases. The term "declared" level may be a tad of a misnomer when we curate.   
That's something to bring up for discussion  too in the curation meeting on Monday alright!. 

When just looking at scans, this would be the one that found in top level key files and package manifests: that really the "declared" thing.
When we curate this should cover the complete licensing IMHO.

Here in anycase,  since all these are in a top level LICENSE and effectively bundled in the project, these would part of the declared license reported by a scan and should also be reported when we curate?

AE49:
@pombredanne Thanks for raising these points and I agree.

ariel11:
The "Declared" license is the overall projejct license, as defined here:  https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md

ariel11:
I do wonder if per our definition of "Declared," we would put MIT **in that field**, but leave all the licenses the tooling has identified for the License file itself....
image

AE49:
@royaljust for visibility

PR #2217, opened 2019-07-18
Correcting license

ariel11:
**Type:** Incorrect

**Summary:**
Correcting license

**Details:**
The overall project license is MIT

**Resolution:**
See: https://github.com/Tencent/rapidjson/blob/17254e090e0dc3d5d1aca8efd6303cdbd07dbae1/license.txt

**Affected definitions**:
- [rapidjson 17254e090e0dc3d5d1aca8efd6303cdbd07dbae1](https://clearlydefined.io/definitions/git/github/tencent/rapidjson/17254e090e0dc3d5d1aca8efd6303cdbd07dbae1/17254e090e0dc3d5d1aca8efd6303cdbd07dbae1)

### DIFF
--- a/curations/git/github/tencent/rapidjson.yaml
+++ b/curations/git/github/tencent/rapidjson.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: rapidjson
+  namespace: tencent
+  provider: github
+  type: git
+revisions:
+  17254e090e0dc3d5d1aca8efd6303cdbd07dbae1:
+    licensed:
+      declared: MIT


ariel11:
@pombredanne - I think you wanted a head's up when the scanners incorrectly identified a license?  It looks to me like that happened here. Scanners identified the declared license as ISC.  It looks like MIT to me (with other licenses applicable to portions of the project, as listed in 'discovered licenses').

pombredanne:
@ariel11 that's an interesting one with a conflict:
1. this is an NPM with effectively a declared ISC license https://github.com/Tencent/rapidjson/blob/17254e090e0dc3d5d1aca8efd6303cdbd07dbae1/package.json#L19
2. per this https://raw.githubusercontent.com/Tencent/rapidjson/17254e090e0dc3d5d1aca8efd6303cdbd07dbae1/license.txt the license is `MIT AND JSON AND BSD-3-Clause`

BUT .... that JSON one is an odd ball and its JSON license only applies to the jsonchecker code that would typically not be part of the main build: it applies only to tests files in https://github.com/Tencent/rapidjson/tree/master/bin/jsonchecker ... So I would not included it in the Declared license.

As for the BSD license the code is bundled and vendored and built in the main  package but ONLY on Windows. https://github.com/Tencent/rapidjson/tree/master/include/rapidjson/msinttypes (this is familiar because I had posted this long ago: https://code.google.com/archive/p/msinttypes/issues/7 ). I would still include it at all times for sanity.

So `MIT AND BSD-3-Clause` would likely be the best declared license.

The other issue from the `ISC` in the package.json: may be @miloyip can help there?  (NB: I verified that this issue still exist (MIT vs. ISC)  in the latest master)
IMHO I think that ISC is a mistake.

pombredanne:
@ariel11 just to clarify in this case: you wrote:

> Scanners identified the declared license as ISC. 

Here the ISC license is effectively present in the package.json and correctly picked up by ScanCode AND I think that we should look not only at package manifests but also at other key files (here the license.txt) when populating the CD "Declared" license field. (this is separate from the curation specifics discussed above)

ariel11:
@pombredanne - this PR is not to curate the license for the NPM....just the license for this GitHub commit (17254e090e0dc3d5d1aca8efd6303cdbd07dbae1). For the GitHub commit, I would go with the license.txt, not the package.json file.  Also, with the 'declared' and 'discovered' license distinction, I thought we were viewing the declared license as the **primary** license(s) that apply to the whole project (except for any files that have other/different copyright/license info).  Per the license.txt, the primary license is MIT and only certain files/dependencies are under the other licenses - "If you have downloaded a copy of the RapidJSON source code from Tencent, please note that RapidJSON source code is licensed under the MIT License, except for the third-party components listed below which are subject to different license terms. ."  So, I think the "declared" license for this GitHub commit would just be "MIT."  Thoughts?

PR #2228, opened 2019-07-25
Add MIT

royaljust:
**Type:** Missing

**Summary:**
Add MIT

**Details:**
MIT license pointed to in nuspec file both in package and at source location.

**Resolution:**
Add MIT license.

**Affected definitions**:
- [knockout.mapping 35482d03ee7520b1afe0b437a6e66150369378a7](https://clearlydefined.io/definitions/git/github/stevesanderson/knockout.mapping/35482d03ee7520b1afe0b437a6e66150369378a7/35482d03ee7520b1afe0b437a6e66150369378a7)

### DIFF
--- a/curations/git/github/stevesanderson/knockout.mapping.yaml
+++ b/curations/git/github/stevesanderson/knockout.mapping.yaml
@@ -4,6 +4,7 @@ coordinates:
   provider: github
   type: git
 revisions:
+  35482d03ee7520b1afe0b437a6e66150369378a7:
     files:
       - license: MIT
         path: Package.nuspec
--- a/curations/git/github/stevesanderson/knockout.mapping.yaml
+++ b/curations/git/github/stevesanderson/knockout.mapping.yaml
@@ -1,0 +1,12 @@
+coordinates:
+  name: knockout.mapping
+  namespace: stevesanderson
+  provider: github
+  type: git
+revisions:
+  35482d03ee7520b1afe0b437a6e66150369378a7:
+    files:
+      - license: MIT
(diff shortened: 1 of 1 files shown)


ariel11:
@pombredanne - I just left a message on Discord about this and here is the PR that inspired the question.  For this commit, there is not a license file in the typical location and there is no license info in the readme.  The nuspec says MIT.  And this says MIT - https://github.com/SteveSanderson/knockout.mapping/blob/35482d03ee7520b1afe0b437a6e66150369378a7/build/version-header.js.  So, I think I am good curating MIT.  But @royaljust and I were wondering what your thoughts in general were on using the nuspec file to curate the 'declared' license for a repo (when the repo doenst have a license file and the readme doesnt indicate a license)?

pombredanne:
NuSpec is the way to go when there is no other or no other conflicting data. 

As a side note in the case here, this JS package is "built" e.g. minified to a more compact using the Google Closure compiler and during this project the https://github.com/SteveSanderson/knockout.mapping/blob/35482d03ee7520b1afe0b437a6e66150369378a7/build/version-header.js is appended to the minified/built JS. This is a tad uncommon way to do things especially for a JS package, but that's OK.

See https://github.com/SteveSanderson/knockout.mapping/blob/35482d03ee7520b1afe0b437a6e66150369378a7/build/build-linux#L11

ariel11:
@geneh - I tried to resolve a conflict and did something.  Can you please assist with this failed check?

geneh:
Done!

PR #2286, opened 2019-07-26
Adding missing BSD3 for this version

carlbrochu:
**Type:** Missing

**Summary:**
Adding missing BSD3 for this version

**Details:**
https://mvnrepository.com/artifact/junit/junit

**Resolution:**
Missing

**Affected definitions**:
- [junit 3.8.1](https://clearlydefined.io/definitions/maven/mavencentral/junit/junit/3.8.1)

### DIFF
--- a/curations/maven/mavencentral/junit/junit.yaml
+++ b/curations/maven/mavencentral/junit/junit.yaml
@@ -6,7 +6,7 @@ coordinates:
 revisions:
   3.8.1:
     licensed:
-      declared: BSD-3-Clause
+      declared: CPL-1.0
   '4.11':
     licensed:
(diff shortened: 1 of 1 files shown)


ariel11:
@carlbrochu - Per the POM, it looks like the license is Common Public License Version 1.0 vs. BSD-3-clause.  Agree?

carlbrochu:
> @carlbrochu - Per the POM, it looks like the license is Common Public License Version 1.0 vs. BSD-3-clause. Agree?

That sounds fine, i tried to extrapolate from the mvn repo and that seems to be flawed.  Since the other version are CPL as well, that seems good to use

PR #2291, opened 2019-07-26
Missing declared

ariel11:
**Type:** Missing

**Summary:**
Missing declared

**Details:**
Followed source link.  "Copyright" file is MIT.

**Resolution:**
Declared MIT

**Affected definitions**:
- [libxml2 4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017](https://clearlydefined.io/definitions/git/github/gnome/libxml2/4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017/4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017)

### DIFF
--- a/curations/git/github/gnome/libxml2.yaml
+++ b/curations/git/github/gnome/libxml2.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017:
     licensed:
-      declared: MIT
+      declared: X11
--- a/curations/git/github/gnome/libxml2.yaml
+++ b/curations/git/github/gnome/libxml2.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: libxml2
+  namespace: gnome
+  provider: github
+  type: git
+revisions:
+  4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017:
+    licensed:
+      declared: MIT


AE49:
@ariel11 should this be "Other" for Declared License because of the extra text added to the MIT license?

curations/git/github/gnome/libxml2.yaml
```diff
@@ -0,0 +1,9 @@
+coordinates:
+  name: libxml2
+  namespace: gnome
+  provider: github
+  type: git
+revisions:
+  4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017:
+    licensed:
+      declared: MIT
```
pombredanne: @ariel11 I think the declared license in the `Copyright` is `X11` not `MIT` 
See https://raw.githubusercontent.com/GNOME/libxml2/4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017/Copyright

This is detected exactly as such by ScanCode at least.

pombredanne:
@AE49 

> should this be "Other" for Declared License because of the extra text added to the MIT license?

In this case, I think there is no such need as this does not introduces any new terms:

https://github.com/GNOME/libxml2/blob/4c2e7c651f6c2f0d1a74f350cbda95f7df3e7017/Copyright

> Except where otherwise noted in the source code (e.g. the files hash.c,
list.c and the trio files, which are covered by a similar licence but
with different Copyright notices) all the files are:


This is more an intro text saying more or less `this is the declared license...`

curations/git/github/gnome/libxml2.yaml
pombredanne: Thanks!

ariel11: Yes, I agree.  I updated the .yaml to X11.

PR #2417, opened 2019-08-29
Declared License

ariel11:
**Type:** Incomplete

**Summary:**
Declared License

**Details:**
Copied the (deprecated) SPDX license identifier given here: https://github.com/ms-iot/u-boot/tree/abec1328dd13df76b19f5fd5ae3a5f3275110c8c/Licenses

**Resolution:**
In the ReadMe: 
SPDX-License-Identifier: GPL-2.0

**Affected definitions**:
- [u-boot abec1328dd13df76b19f5fd5ae3a5f3275110c8c](https://clearlydefined.io/definitions/git/github/ms-iot/u-boot/abec1328dd13df76b19f5fd5ae3a5f3275110c8c/abec1328dd13df76b19f5fd5ae3a5f3275110c8c)

### DIFF
--- a/curations/git/github/ms-iot/u-boot.yaml
+++ b/curations/git/github/ms-iot/u-boot.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   abec1328dd13df76b19f5fd5ae3a5f3275110c8c:
     licensed:
-      declared: GPL-2.0
+      declared: GPL-2.0-only
--- a/curations/git/github/ms-iot/u-boot.yaml
+++ b/curations/git/github/ms-iot/u-boot.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: u-boot
+  namespace: ms-iot
+  provider: github
+  type: git
+revisions:
+  abec1328dd13df76b19f5fd5ae3a5f3275110c8c:
+    licensed:
+      declared: GPL-2.0


AE49:
@pombredanne can we get your thoughts on this curation? @ariel11 copied the deprecated SPDX license and is wondering if we should keep the deprecated identifiers in the license option list.

pombredanne:
We should always use the current latest version of ids. This is also what the latest scancode version does. Here we should use GPL-2.0-Only and not some legacy ids IMHO.

ariel11:
Thanks @pombredanne and @AE49.  Given this https://github.com/ms-iot/u-boot/blob/abec1328dd13df76b19f5fd5ae3a5f3275110c8c/Licenses/Exceptions, I wonder if the "declared" field should actually be "GPL-2.0-only AND GPL-2.0-or later," since it says "U-Boot in general is covered by the GPL-2.0/GPL-2.0+."

ariel11:
@AE49 - thoughts on the above?

royaljust:
This further highlights the need for some consistency about how we think about our constructs of "declared" versus "discovered" licenses.  In this case, I'd go with GPL-2.0-only because, as a whole, this is the only way the component can be distributed, even though the majority of files might be GPL-2.0-or-later.

ariel11:
Thanks, @royaljust.  I totally agree.  @AE49 - I updated the yaml so this should be ready to approve/merge.  Thanks everyone!

PR #2474, opened 2019-09-10
Delcaring license

ariel11:
**Type:** Incomplete

**Summary:**
Delcaring license

**Details:**
Followed source link to license file

**Resolution:**
MIT

**Affected definitions**:
- [jsoncpp 645250b6690785be60ab6780ce4b58698d884d11](https://clearlydefined.io/definitions/git/github/open-source-parsers/jsoncpp/645250b6690785be60ab6780ce4b58698d884d11)

### DIFF
--- a/curations/git/github/open-source-parsers/jsoncpp.yaml
+++ b/curations/git/github/open-source-parsers/jsoncpp.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: jsoncpp
+  namespace: open-source-parsers
+  provider: github
+  type: git
+revisions:
+  645250b6690785be60ab6780ce4b58698d884d11:
+    licensed:
+      declared: MIT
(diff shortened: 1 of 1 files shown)


capfei:
@ariel11, correct me if I am wrong since I've been out of this a while, but should it be NO ASSERTION OR MIT OR NO ASSERTION AND MIT? I could be reading the jurisdiction parts incorrectly.

ariel11:
@capfei - we have added "NONE" and "OTHER" as curation options.  We have been doing "OTHER" for public domain.  Looking at the https://github.com/open-source-parsers/jsoncpp/blob/645250b6690785be60ab6780ce4b58698d884d11/LICENSE file again, I don't think we can do "MIT AND/OR OTHER," which would most closely match the three licensing options in the license file.  I suppose I would recommend putting "MIT OR OTHER" as the curation here.  Thoughts?  Good catch

capfei:
Ok, I will update the PR to MIT OR OTHER. Thanks!

PR #2488, opened 2019-09-13
MediaTypeMap.Core

LTuga:
**Type:** Missing

**Summary:**
MediaTypeMap.Core

**Details:**
Entry missing.

**Resolution:**
Added source code repository found on associated NuGet gallery page:
https://www.nuget.org/packages/MediaTypeMap.Core/

Added license type found in source code repository.
https://github.com/samuelneff/MimeTypeMap

Not sure why author has differing names between NuGet and source repo.

**Affected definitions**:
- [MediaTypeMap.Core 2.3.3](https://clearlydefined.io/definitions/nuget/nuget/-/MediaTypeMap.Core/2.3.3)

### DIFF
--- a/curations/nuget/nuget/-/MediaTypeMap.Core.yaml
+++ b/curations/nuget/nuget/-/MediaTypeMap.Core.yaml
@@ -1,0 +1,16 @@
+coordinates:
+  name: MediaTypeMap.Core
+  provider: nuget
+  type: nuget
+revisions:
+  2.3.3:
+    described:
+      sourceLocation:
+        name: mimetypemap
+        namespace: samuelneff
+        provider: github
+        revision: 6ddc67a6ddd873ca720f76847edc34d0c7ee5e06
+        type: git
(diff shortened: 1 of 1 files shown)


capfei:
@LTuga, the commit ID provided doesn't indicate it is for version 2.3.3, but the latest commit in the repo. The commit is from Oct 24, 2018 and the NuGet was published in April 2018. I am not able to figure out which commit is for 2.3.3 since there is no version bumping after 2.3.2 on December 28, 2017. If you can provide the actual commit ID for 2.3.3, I will update the PR, otherwise, we can remove the GitHub source information since I cannot verify which commit is the correct one. The NuGet publish date doesn't appear to coincide with version dates since 2.3.2 is from 2017 but the package is April 2018.

LTuga:
I can't provide the exact commit either, but the Contribute button wouldn't let me list the GitHub repo without a commit hash. My goal was to display the repo for future viewers to correlate in some capacity, but perhaps that is also not helpful if there isn't a comment explaining the author's naming differences between NuGet and GitHub. Feel free to modify as you see fit.

capfei:
Source location has been removed from the PR. Approved and merged.

PR #2574, opened 2019-10-15
Missing Declared

ariel11:
**Type:** Missing

**Summary:**
Missing Declared

**Details:**
GitHub POM indicates license is Apache v2

**Resolution:**
https://github.com/codehaus/jettison/blob/master/pom.xml

**Affected definitions**:
- [jettison 1.1](https://clearlydefined.io/definitions/maven/mavencentral/org.codehaus.jettison/jettison/1.1)


pombredanne:
:+1:  and FWIW, this would also would be a "declared" license in my book as it was also at the time
https://github.com/codehaus/jettison/blob/jettison-1.1/src/main/resources/META-INF/LICENSE

PR #2726, opened 2019-11-07
Missing declared license

capfei:
**Type:** Missing

**Summary:**
Missing declared license

**Details:**
Missing license

**Resolution:**
Went with OTHER since the MIT license includes additional verbiage about requiring a HighCharts license for commercial use. If it should be something else, please update. https://github.com/highcharts/highcharts-react/blob/v2.2.2/LICENSE

**Affected definitions**:
- [highcharts-react-official 2.2.2](https://clearlydefined.io/definitions/npm/npmjs/-/highcharts-react-official/2.2.2)


ariel11:
Agree with OTHER due to the additional language

PR #2931, opened 2019-11-27
Resolve cases of NOASSERTION

bduranc:
**Type:** Incorrect

**Summary:**
Resolve cases of NOASSERTION

**Details:**
All of the files referenced were incorrectly declared as "NOASSERTION".

They are actually all under Amazon Software License (ASL) which is not a valid SPDX license.

See: https://aws.amazon.com/asl/

**Resolution:**
Updated the declared license of all references files in accordance with their content/headers. Because the Amazon Software License is not a valid SPDX license, files were declared as "OTHER".

**Affected definitions**:
- [amazon-kinesis-producer 0.12.11](https://clearlydefined.io/definitions/sourcearchive/mavencentral/com.amazonaws/amazon-kinesis-producer/0.12.11)

### DIFF
--- a/curations/sourcearchive/mavencentral/com.amazonaws/amazon-kinesis-producer.yaml
+++ b/curations/sourcearchive/mavencentral/com.amazonaws/amazon-kinesis-producer.yaml
@@ -1,0 +1,66 @@
+coordinates:
+  name: amazon-kinesis-producer
+  namespace: com.amazonaws
+  provider: mavencentral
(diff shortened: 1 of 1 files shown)


bduranc:
Copyright attributions are being added to scope of curation unintentionally due to issue https://github.com/clearlydefined/service/issues/677. They should be the same as their original values prior to curation, so no harm done in this case.

PR #2993, opened 2019-12-05
Missing declared license

capfei:
**Type:** Missing

**Summary:**
Missing declared license

**Details:**
Missing license

**Resolution:**
Apache 2.0: https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/2.19.208020213/LICENSE

**Affected definitions**:
- [Microsoft.IdentityModel.Clients.ActiveDirectory 2.19.208020213](https://clearlydefined.io/definitions/nuget/nuget/-/Microsoft.IdentityModel.Clients.ActiveDirectory/2.19.208020213/2.19.208020213)

### DIFF
--- a/curations/nuget/nuget/-/Microsoft.IdentityModel.Clients.ActiveDirectory.yaml
+++ b/curations/nuget/nuget/-/Microsoft.IdentityModel.Clients.ActiveDirectory.yaml
@@ -3,6 +3,9 @@ coordinates:
   provider: nuget
   type: nuget
 revisions:
+  2.19.208020213:
+    licensed:
+      declared: Apache-2.0
   2.22.302111727:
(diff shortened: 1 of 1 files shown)


ariel11:
@capfei - the curation guidelines have been updated to say that we would curate the license link on the package page, even if it points to the master branch vs. digging into the source.

See: https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md
"1. Target component metadata has license info indicating a license URL, but the link points to a volatile location. 
Apply license found at the volatile location at the time checked."

So, we would curate MIT (over Apache in the source).

Agreed?

capfei:
Updated PR to MIT

PR #3020, opened 2019-12-09
Resolve case of NOASSERTION

bduranc:
**Type:** Incorrect

**Summary:**
Resolve case of NOASSERTION

**Details:**
License was incorrectly discovered as "BSD-3-Clause AND NOASSERTION."

BSD code is from https://github.com/mikolalysenko/angle

Starting on line # 3052, there is also some GPL-licensed code generated from Bison project.

**Resolution:**
Leave BSD-3-Clause as-is, although header says "BSD-style" license, it can be confirmed here as BSD-3-Clause https://github.com/mikolalysenko/angle/blob/master/LICENSE.

But the NOASSERTION should be changed to "GPL-2.0-with-bison-exception" given discovery of the Bison code.

**Affected definitions**:
- [aports b8229586a20eea5fc0f782a383ef666b6a60e054](https://clearlydefined.io/definitions/git/github/alpinelinux/aports/b8229586a20eea5fc0f782a383ef666b6a60e054)


capfei:
Fixed this; I had `attributions` tabbed over too far after resolving conflict and didn't realize it until now.

PR #3153, opened 2019-12-18
Resolve cases of NOASSERTION

bduranc:
**Type:** Incorrect

**Summary:**
Resolve cases of NOASSERTION

**Details:**
All of the referenced files captured by this curation have "MIT-no-advert-export" license (see: https://enterprise.dejacode.com/licenses/public/mit-no-advert-export-control/#license-text) which is not listed explicitly in SPDX, so "OTHER" is used instead.

**Resolution:**
In addition to the discovery of "MIT-no-advert-export" license mentioned above, some incorrectly discovered BSD findings were also found and corrected (i.e. discovered originally as BSD-2-Clause, but is actually  3-Clause).

**Affected definitions**:
- [krb5 97e3c42b2a89a2ec60eb93d3f974769e3e3cbdc5](https://clearlydefined.io/definitions/git/github/krb5/krb5/97e3c42b2a89a2ec60eb93d3f974769e3e3cbdc5/97e3c42b2a89a2ec60eb93d3f974769e3e3cbdc5)

### DIFF
--- a/curations/git/github/krb5/krb5.yaml
+++ b/curations/git/github/krb5/krb5.yaml
@@ -7,35 +7,39 @@ revisions:
   97e3c42b2a89a2ec60eb93d3f974769e3e3cbdc5:
     files:
       - attributions:
-          - 'Copyright (c) 2006-2008, Novell, Inc. All Rights Reserved.'
-          - Copyright 1990 by the Massachusetts Institute of Technology. All Rights Reserved.
-        license: BSD-3-Clause AND OTHER
-        path: src/lib/krb5/krb/authdata_dec.c
+          - 'Copyright 2013,2014 Red Hat, Inc.'
+          - 'Copyright 1990,1991,2001,2002,2004,2005,2007,2008 by the Massachusetts Institute of Technology.'
+        license: BSD-2-Clause AND OTHER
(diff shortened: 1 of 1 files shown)


bduranc:
I notice that the copyright attributions have been unintentionally included in this contribution again, likely due to pending bug https://github.com/clearlydefined/website/issues/855.

This is not ideal as my curation has unintentionally removed the "All Rights Reserved" portions from the previous attributions. Should I submit another curation to fix this, or can this be fixed pre-merge?

I never touched the attribution fields yet for some reason they "self highlight" for each license SPDX I intended to contribute. I am not presented with an option to revert on my end.

bduranc:
This particular curation was only intended to correct file-level declared license (where discovered) as NOASSERTION) information.

However, it appears copyright attributions have also unintentionally entered the scope of this curation (likely due to open bug https://github.com/clearlydefined/website/issues/855) and I took notice that the tools did not discover or add the "All Rights Reserved" portion of the various copyright statements (I didn't update that field, so it would take whatever has been discovered).

Should I just submit a second curation for copyright attributions to include these portions, or can we make the change here? I does not exist currently in the discoveries.

capfei:
@bduranc I'll fix the PR since I need to review it anyways.

PR #3163, opened 2019-12-19
Declaring license and License file

ariel11:
**Type:** Missing

**Summary:**
Declaring license and License file

**Details:**
Not sure if this should be AGPL-3.0-only or -or-later.  The license file says "..either version 3 of the License," so I think it was supposed to be "either version 3...or later."

**Resolution:**
https://clearlydefined.io/file/304de4d159ae692655bbb038824c7675bbcf10541583b6fe1402198e7332eba0

**Affected definitions**:
- [Newtonsoft.Json.Schema 3.0.11](https://clearlydefined.io/definitions/nuget/nuget/-/Newtonsoft.Json.Schema/3.0.11/3.0.11)

### DIFF
--- a/curations/nuget/nuget/-/Newtonsoft.Json.Schema.yaml
+++ b/curations/nuget/nuget/-/Newtonsoft.Json.Schema.yaml
@@ -8,7 +8,7 @@ revisions:
       declared: AGPL-3.0 OR OTHER
   3.0.11:
     files:
-      - license: AGPL-3.0-or-later
+      - license: AGPL-3.0-or-later OR OTHER
         path: LICENSE.md
     licensed:
-      declared: AGPL-3.0-or-later
+      declared: AGPL-3.0-or-later OR OTHER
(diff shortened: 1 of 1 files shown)


capfei:
@ariel11 I agree that that phrase indicates AGPL-3.0-or-later. It looks like it is also available under a commercial license. Would you like me to update to AGPL-3.0-or-later or OTHER?

ariel11:
@capfei Thank you.  That's an interesting question.  I saw the note about the commercial license, though I was thinking of those terms as separately available and not something that applies to this code now.  In other words, the commercial license is an _optional_ "other" vs. there being other license terms that apply to this code now, so I left it out of the curation. Thoughts?  What do you think @AE49 or @royaljust?

capfei:
Unassigning myself since I will be out for the rest of the year.

ariel11:
@capfei - Thinking about this again, I agree with AGPL-3.0-or-later or OTHER.  Please update and review. Thanks!

PR #3262, opened 2020-01-06
Tech Team Scan Update

clearlyoss:
**Type:** Incomplete

**Summary:**
Tech Team Scan Update

**Details:**
The section of code following this notice originates with, or is related to, WebSockets, an "[i]mplementation of the WebSocket protocol, along with client and server integration components." See https://github.com/aspnet/WebSockets/. This material is licensed under the Apache Software License v2.0 

**Resolution:**
Fixes license and copyright attribution

**Affected definitions**:
- [azure-relay-dotnet 13ff04770abe8ca4e14cc6d58f00d0d9d712437d](https://clearlydefined.io/definitions/git/github/azure/azure-relay-dotnet/13ff04770abe8ca4e14cc6d58f00d0d9d712437d/13ff04770abe8ca4e14cc6d58f00d0d9d712437d)

### DIFF
--- a/curations/git/github/azure/azure-relay-dotnet.yaml
+++ b/curations/git/github/azure/azure-relay-dotnet.yaml
@@ -1,0 +1,12 @@
+coordinates:
+  name: azure-relay-dotnet
+  namespace: azure
+  provider: github
+  type: git
+revisions:
+  13ff04770abe8ca4e14cc6d58f00d0d9d712437d:
+    files:
+      - attributions:
+          - '  Copyright (c) .NET Foundation and Contributors'
+        license: Apache-2.0
+        path: src/Microsoft.Azure.Relay/WebSockets/NetStandard20/ManagedWebSocket.cs


capfei:
@clearlyoss the file points to https://github.com/dotnet/corefx/blob/master/src/Common/src/System/Net/WebSockets/ManagedWebSocket.cs, which is MIT. When I search aspnet/WebSockets history, the file says MIT. See https://github.com/aspnet/WebSockets/blob/release/2.0/src/Microsoft.AspNetCore.WebSockets/Internal/fx/src/Common/src/System/Net/WebSockets/ManagedWebSocket.cs. Other than the license file in aspnet/WebSockets, is there somewhere else that this file states its Apache?

clearlyoss:
Hello - take a look below:

File   containing a section of code prefaced by the following notice:                  // From   https://raw.githubusercontent.com/aspnet/WebSockets/dev/src/Microsoft.AspNetCore.WebSockets.Protocol/Utilities.cs             // Performs a stateful   validation of UTF-8 bytes.             // It checks for valid   formatting, overlong encodings, surrogates, and value ranges.          The section of code following this notice originates with, or is related   to, WebSockets, an "[i]mplementation of the WebSocket protocol, along   with client and server integration components." See   https://github.com/aspnet/WebSockets/. This material is licensed under the   Apache Software License v2.0 (see   https://github.com/aspnet/WebSockets/blob/master/LICENSE.txt);   azure-relay-dotnet #13ff04770abe8ca4e14cc6d58f00d0d9d712437d may not be in   compliance with the terms of this license.
--

capfei:
The link doesn't work for me and that is a different file. ManagedWebSocket.cs states that it is a shared file with dotnet/corefx, which is MIT. https://github.com/aspnet/WebSockets/blob/release/2.0/src/Microsoft.AspNetCore.WebSockets/Internal/fx/src/Common/src/System/Net/WebSockets/ManagedWebSocket.cs#L16. 

It also says at the top that it is MIT: 
`// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.`

When I go up in the directory, the readme file that states that the code under src folder is from dotnet/corefx. https://github.com/aspnet/WebSockets/tree/release/2.0/src/Microsoft.AspNetCore.WebSockets/Internal/fx

clearlyoss:
Hi Candice,
This in-file comment states there is Apache licensed code inside this otherwise MIT file. The comment in the file which states this section of code came from an upstream Apache licensed project, so it would be appropriate to attribute the upstream project and pass along the Apache license as well. So there is no misunderstanding somebody added Apache material to this file that originated under MIT, and I'm not trying to change the overall license MIT - just add this attribution.  Hope that makes sense ;)

capfei:
Ok, I think I see. So it should be `MIT AND APACHE-2.0` to cover the overall MIT license and the portion that contains Apache, correct?

pombredanne:
Just to be sure, we rea talking about this file https://github.com/Azure/azure-relay-dotnet/blob/13ff04770abe8ca4e14cc6d58f00d0d9d712437d/src/Microsoft.Azure.Relay/WebSockets/NetStandard20/ManagedWebSocket.cs correct?

pombredanne:
I get the Apache reference now. IMHO that's fine to put this as Apache too and also reach out to the authors upstream to clean things up.
For the sake of clarity we should also add some summary write up that explains very clearly the trail we followed.

clearlyoss:
> Ok, I think I see. So it should be `MIT AND APACHE-2.0` to cover the overall MIT license and the portion that contains Apache, correct?

I think we are closed and good on this issue but re: the above question no - that would be dual licensed.  This would be adding and addition attribution for Apache though the overall license is still MIT.

PR #3331, opened 2020-01-11
Declaring license

ariel11:
**Type:** Missing

**Summary:**
Declaring license

**Details:**
NuGet license link 

**Resolution:**
https://raw.githubusercontent.com/OData/odata.net/master/LICENSE.txt

**Affected definitions**:
- [Microsoft.OData.Edm 7.6.0](https://clearlydefined.io/definitions/nuget/nuget/-/Microsoft.OData.Edm/7.6.0/7.6.0)

### DIFF
--- a/curations/nuget/nuget/-/Microsoft.OData.Edm.yaml
+++ b/curations/nuget/nuget/-/Microsoft.OData.Edm.yaml
@@ -26,4 +26,4 @@ revisions:
       declared: MIT
   7.6.0:
     licensed:
-      declared: MIT AND CC-BY-3.0
+      declared: MIT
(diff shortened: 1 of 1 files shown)


ariel11:
Debated whether to do just MIT or MIT AND CC BY 3.0.  Both apply to the _repo_, though only MIT applies to the code. We should add what we decide to the trouble shooting section of the curation guidelines.

capfei:
My first reaction is to say MIT since the NuGet doesn't contain any documentation, only code. I'll leave this open for a bit longer in case others want to provide their thoughts on this situation.

capfei:
Adding a tag to help make it stand out

ariel11:
@capfei - that makes sense.  And I think just MIT is consistent on other versions.   I am good with that.

PR #3441, opened 2020-01-22
Correcting Declared license

ariel11:
**Type:** Incorrect

**Summary:**
Correcting Declared license

**Details:**
The package metadata says "GNU Library or Lesser General Public License (LGPL) (LGPL)," so LGPL-2.0 or 2.1.•	The setup.py says LGPL-2.1 or later https://github.com/paramiko/paramiko/blob/690d142321f53ae82f49a57b7f6716fb947d5687/setup.py.

**Resolution:**
So, I opted for LGPL-2.0-or later.

**Affected definitions**:
- [paramiko 2.7.1](https://clearlydefined.io/definitions/pypi/pypi/-/paramiko/2.7.1/2.7.1)

### DIFF
--- a/curations/pypi/pypi/-/paramiko.yaml
+++ b/curations/pypi/pypi/-/paramiko.yaml
@@ -8,4 +8,4 @@ revisions:
       declared: LGPL-2.1-only
   2.7.1:
     licensed:
-      declared: LGPL-2.0-or-later
+      declared: LGPL-2.1-or-later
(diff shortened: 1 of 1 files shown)


capfei:
I would say it should be 2.1 or later since the license file included with package contents is 2.1 only and the headers say 2.1 or later. Let me know what you think.

ariel11:
@royaljust and @capfei  - well, the PR to our curation guidelines does say to prioritize the license file over the metadata to the extent there is a conflict.  Though here, I dont know if we necessarily have a "conflict."  The metadata mentions "Library," which is the LGPL-2.0 license and LGPL-2.0-or later also encompasses LGPL-2.1.  Thoughts?

ariel11:
@capfei - OK, I am good with going with the license file and doing LGPL-2.1-or later. :)

PR #3463, opened 2020-02-03
Add Declared License

AE49:
**Type:** Incomplete

**Summary:**
Add Declared License

**Details:**
Adding MIT and Other for Highcharts Proprietary License

**Resolution:**
ClearlyDefined license: MIT and Other(https://clearlydefined.io/file/e967614a786bb60fb60eb29b79d8554b1ad768db3d61379798d8142175822944)

GitHub project site license: https://github.com/highcharts/highcharts-react/blob/v2.1.3/LICENSE

**Affected definitions**:
- [highcharts-react-official 2.1.3](https://clearlydefined.io/definitions/npm/npmjs/-/highcharts-react-official/2.1.3/2.1.3)

### DIFF
--- a/curations/npm/npmjs/-/highcharts-react-official.yaml
+++ b/curations/npm/npmjs/-/highcharts-react-official.yaml
@@ -8,4 +8,4 @@ revisions:
       declared: MIT AND OTHER
   2.2.2:
     licensed:
-      declared: OTHER
+      declared: MIT AND OTHER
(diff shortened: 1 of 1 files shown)


pombredanne:
that's an interesting case of what I call OINA ...  "Open source in name only".  :slight_smile:
Practically the code is useless unless you have an API key of sorts and a commercial contract. So it essentially implies that there should be such a commercial license in place.

1. I agree with @KPFleming (from a chat) that this commercial dep should not affect the license and probably does not affect it here

2. we still would want to have a way to signal that there are dependencies that have different terms. It boils down to the requirement from @duranbc (Brian Duran) e.g. track deps too

OINA are common enough and typically problematic. So It likely makes sense to report `AND OTHER` as you did here as a way to signal this until we have a way to deal with deps.

capfei:
Thank you, @pombredanne!

PR #3646, opened 2020-03-03
Correct ring expression

Jake-Shadle:
**Type:** Ambiguous

**Summary:**
Correct ring expression

**Details:**
The ring project has very complicated licensing due to its ancestry, but it is clearly stated in the LICENSE file in the project itself

**Resolution:**
This PR applies my understanding of the licensing terms spelled out in the top of the LICENSE file into the (I believe) correct top-level SPDX expression for the project

**Affected definitions**:
- [ring 0.16.11](https://clearlydefined.io/definitions/crate/cratesio/-/ring/0.16.11)

### DIFF
--- a/curations/crate/cratesio/-/ring.yaml
+++ b/curations/crate/cratesio/-/ring.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: ring
+  provider: cratesio
+  type: crate
+revisions:
+  0.16.11:
+    licensed:
+      declared: ISC AND OpenSSL AND MIT


Jake-Shadle:
More details from the LICENSE file.

> *ring* uses an ISC-style license like BoringSSL for code in new files, including in particular all the Rust code

> BoringSSL is a fork of OpenSSL. As such, large parts of it fall under OpenSSL licensing. Files that are completely new have a Google copyright and an ISC license. This license is reproduced at the bottom of this file.

> Files in third_party/ have their own licenses, as described therein. The MIT license, for third_party/fiat, which, unlike other third_party directories, is compiled into non-test libraries, is included below.

> The OpenSSL toolkit stays under a dual license, i.e. both the conditions of the OpenSSL License and the original SSLeay license apply to the toolkit. See below for the actual license texts.

AFAIK, as far as SPDX is concernced, the OpenSSL license is the singular correct license identifier for the "OpenSSL" license, which includes the "SSLeay" license.

PR #3868, opened 2020-04-21
Add Declared License

AE49:
**Type:** Incomplete

**Summary:**
Add Declared License

**Details:**
Adding MIT

**Resolution:**
Package license link goes to CodePlex: https://archive.codeplex.com/?p=aspnetwebstack, which further points to GitHub: https://github.com/OData/WebApi/blob/master/License.txt

**Affected definitions**:
- [Microsoft.AspNet.Mvc 5.1.0](https://clearlydefined.io/definitions/nuget/nuget/-/Microsoft.AspNet.Mvc/5.1.0/5.1.0)

### DIFF
--- a/curations/nuget/nuget/-/Microsoft.AspNet.Mvc.yaml
+++ b/curations/nuget/nuget/-/Microsoft.AspNet.Mvc.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: Microsoft.AspNet.Mvc
+  provider: nuget
+  type: nuget
+revisions:
+  5.1.0:
+    licensed:
+      declared: MIT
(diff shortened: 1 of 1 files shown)


capfei:
@ae49 Newer versions point the source to https://github.com/aspnet/AspNetWebStack. The source repo is Apache 2.0, but most of the NuGet packages list a EULA for the license, including the latest.

https://www.nuget.org/packages/Microsoft.AspNet.Mvc/5.2.7

I think this should be OTHER. If you agree, I will update the PR.

AE49:
@capfei I see your findings with the  Apache 2.0 in the GitHub repo and the Microsoft EULA for the later versions. I would agree with you, it makes sense to curate this version 5.1.0 as Other for the NuGet package. I made the edit. Thanks.

PR #4010, opened 2020-05-08
Declared

ariel11:
**Type:** Missing

**Summary:**
Declared

**Details:**
Nothing in “Files” section of ClearlyDefined
No source link
NuGet “license” link says “Apache-2.0 License” and goes to CodePlex archive
Navigate to GitHub, version is Apache v2 - https://github.com/OData/WebApi/blob/v3.1/License.txt

**Resolution:**
The license today is MIT, but for this version, it is Apache-2.0 and copyright MS Open Tech

**Affected definitions**:
- [Microsoft.AspNet.WebPages 3.1.0](https://clearlydefined.io/definitions/nuget/nuget/-/Microsoft.AspNet.WebPages/3.1.0/3.1.0)

### DIFF
--- a/curations/nuget/nuget/-/Microsoft.AspNet.WebPages.yaml
+++ b/curations/nuget/nuget/-/Microsoft.AspNet.WebPages.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: Microsoft.AspNet.WebPages
+  provider: nuget
+  type: nuget
+revisions:
+  3.1.0:
+    licensed:
+      declared: Apache-2.0


capfei:
@ariel11 Just an FYI, later versions of this NuGet point to the source as https://github.com/aspnet/AspNetWebStack, which is still Apache 2.0.

https://github.com/aspnet/AspNetWebStack/blob/master/LICENSE.txt

ariel11:
Thanks @capfei.  That is interesting.  The path I followed was the NuGet "license info" link takes you here: https://archive.codeplex.com/?p=aspnetwebstack.  On that page it says the Web API OData  project is now here https://github.com/odata/webapi and that is MIT (https://github.com/OData/WebApi/blob/master/License.txt).  Looks like you found a different repo. 
 Also, interesting - now that I look, the _latest_ NuGet is under a EULA - https://www.nuget.org/packages/Microsoft.AspNet.WebPages/3.2.7.  However, I think we are in agreement this particular NuGet is under Apache v2. :)

capfei:
@ariel11 They way I read it is that source for the three NuGet packages (MVC, WebAPI, and Web Pages) used to reside all together on CodePlex, but that WebAPI split off separately to OData while MVC and Web Pages stayed under aspnet. This could be helpful with different versions of these packages.

Too bad there isn't a way to make a note on the curation site so others could see it.

PR #4175, opened 2020-05-21
Declared License

AE49:
**Type:** Incomplete

**Summary:**
Declared License

**Details:**
Adding AGPL v3.0 and Other for additional commercial license

**Resolution:**
NuGet license: https://raw.githubusercontent.com/JamesNK/Newtonsoft.Json.Schema/master/LICENSE.md

Source license: https://github.com/JamesNK/Newtonsoft.Json.Schema/blob/17c43722ab31ed5e50c1c230dd1d7f7d9cb7eee2/LICENSE.md

**Affected definitions**:
- [Newtonsoft.Json.Schema 2.0.8](https://clearlydefined.io/definitions/nuget/nuget/-/Newtonsoft.Json.Schema/2.0.8/2.0.8)

### DIFF
--- a/curations/nuget/nuget/-/Newtonsoft.Json.Schema.yaml
+++ b/curations/nuget/nuget/-/Newtonsoft.Json.Schema.yaml
@@ -9,6 +9,9 @@ revisions:
   2.0.2:
     licensed:
       declared: AGPL-3.0-or-later OR OTHER
+  2.0.8:
+    licensed:
+      declared: AGPL-3.0 AND OTHER
   3.0.10:
     licensed:
       declared: AGPL-3.0-or-later OR OTHER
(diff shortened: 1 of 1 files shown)


capfei:
@ae49 I'm adding this to the agenda so I can get others opinion on if this should be 3.0-only or 3.0-or-later since it has the verbiage, "either version 3 of the License" that is in 3.0-or-later, but it doesn't include, "or (at your option) any later version."

capfei:
Opened an issue to find out which identifier it should be.
https://github.com/JamesNK/Newtonsoft.Json.Schema/issues/218

capfei:
Author confirmed AGPL-3.0-only.

ariel11:
@capfei - Per the author, its "AGPL-3.0-only."  We'll need to update the ones curated with "or-later."

PR #4407, opened 2020-06-17
Pillow 7.1.2

AE49:
**Type:** Incomplete

**Summary:**
Pillow 7.1.2

**Details:**
Declaring Other for Python Imaging Library License (PIL). 

**Resolution:**
I don't think there is an SPDX identifier for this license.

https://pillow.readthedocs.io/en/stable/about.html#license

https://github.com/python-pillow/Pillow/blob/7.1.2/LICENSE

**Affected definitions**:
- [Pillow 7.1.2](https://clearlydefined.io/definitions/pypi/pypi/-/Pillow/7.1.2/7.1.2)

### DIFF
--- a/curations/pypi/pypi/-/Pillow.yaml
+++ b/curations/pypi/pypi/-/Pillow.yaml
@@ -27,3 +27,6 @@ revisions:
   7.1.1:
     licensed:
       declared: HPND
+  7.1.2:
+    licensed:
+      declared: OTHER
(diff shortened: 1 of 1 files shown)


capfei:
@AE49 This is has been curated as HPND since that is what is listed in the package metadata. However, I have come across related conversations that state the [PIL license](https://github.com/python-pillow/Pillow/blob/master/LICENSE) is a match to [MIT-CMU](https://spdx.org/licenses/MIT-CMU.html) instead of [HPND](https://spdx.org/licenses/HPND.html). 
https://github.com/spdx/license-list-XML/issues/907

I'm reaching out the community to find out more information about these past conversations and if there was a definite resolution to the which identifier it should be.

capfei:
From what I understand, pypi currently doesn't support MIT-CMU as a license indicator so I will process as-is.

PR #4469, opened 2020-06-22
pywin32	228

mihouc:
**Type:** Missing

**Summary:**
pywin32	228

**Details:**
License file in package indicates license is BSD-3

**Resolution:**
While the PyPi site indicates the license is Python Software Foundation License, the actual license file in the package is BSD-3

**Affected definitions**:
- [pywin32 228](https://clearlydefined.io/definitions/pypi/pypi/-/pywin32/228/228)

### DIFF
--- a/curations/pypi/pypi/-/pywin32.yaml
+++ b/curations/pypi/pypi/-/pywin32.yaml
@@ -15,3 +15,6 @@ revisions:
   '227':
     licensed:
       declared: Python-2.0
+  '228':
+    licensed:
+      declared: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


capfei:
@mihouc It looks like each folder has a license, most the same but two are different. The package metadata and the `README` in the source repo states the license for pywin32 is `PSF-2.0`: https://github.com/mhammond/pywin32/blob/master/README.md.

Do you agree it should be curated as `Python-2.0`?

mihouc:
Yes. Thanks

PR #4522, opened 2020-06-23
Add Other License

TRQ3:
**Type:** Missing

**Summary:**
Add Other License

**Details:**
No info in package files. Meta data points to Bouncy Castle License. Curated as Other. 

**Resolution:**
http://www.bouncycastle.org/licence.html

**Affected definitions**:
- [BouncyCastle 1.8.3.1](https://clearlydefined.io/definitions/nuget/nuget/-/BouncyCastle/1.8.3.1/1.8.3.1)

### DIFF
--- a/curations/nuget/nuget/-/BouncyCastle.yaml
+++ b/curations/nuget/nuget/-/BouncyCastle.yaml
@@ -12,6 +12,9 @@ revisions:
   1.8.2:
     licensed:
       declared: MIT
+  1.8.3.1:
+    licensed:
+      declared: OTHER
   1.8.4:
     licensed:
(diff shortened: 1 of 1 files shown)


capfei:
@TRQ3 The license is MIT. Why do you think it should be curated as OTHER?

TRQ3:
@capfei We had Bouncy Castle come up before, we reviewed on our call, and my notes said curate as Other. I see that is references MIT and the language aligns. I just remember this one because of the funny name. If you think we should curate as MIT, that's fine too.

capfei:
@TRQ3 Thanks! I updated to MIT.

PR #4621, opened 2020-06-29
net-dns 0.8.0

AE49:
**Type:** Incomplete

**Summary:**
net-dns 0.8.0

**Details:**
Declaring Ruby

**Resolution:**
ClearlyDefined Files: README.md states Ruby

https://github.com/bluemonk/net-dns/tree/v0.8.0

**Affected definitions**:
- [net-dns 0.8.0](https://clearlydefined.io/definitions/gem/rubygems/-/net-dns/0.8.0/0.8.0)

### DIFF
--- a/curations/gem/rubygems/-/net-dns.yaml
+++ b/curations/gem/rubygems/-/net-dns.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: net-dns
+  provider: rubygems
+  type: gem
+revisions:
+  0.8.0:
+    licensed:
+      declared: Ruby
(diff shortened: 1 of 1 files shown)


ariel11:
@AE49 - Hi, the README says this is licensed under the same license as Ruby is.  Ruby is licensed under BSD-2-Cluase OR Ruby.  I think that should be the curation here.  Agree?

capfei:
I agree with @ariel11 that it should be `BSD-2-Clause or Ruby` since it does state it is the same as Ruby. Ruby has been dual licensed with BSD-2-Clause since 2010.
https://github.com/ruby/ruby/commit/2cd6800fd8437b1f862f3f5c44db877159271d17
https://www.ruby-lang.org/en/about/license.txt

AE49:
@ariel11 I made the edit to the file. Thanks.

PR #4721, opened 2020-07-02
Newtonsoft.Json.Schema 1.0.11

TRQ3:
**Type:** Missing

**Summary:**
Newtonsoft.Json.Schema 1.0.11

**Details:**
No info in package files. Meta data confirms AGPL 3.0. 

**Resolution:**
https://raw.githubusercontent.com/JamesNK/Newtonsoft.Json.Schema/master/LICENSE.md

**Affected definitions**:
- [Newtonsoft.Json.Schema 1.0.11](https://clearlydefined.io/definitions/nuget/nuget/-/Newtonsoft.Json.Schema/1.0.11/1.0.11)

### DIFF
--- a/curations/nuget/nuget/-/Newtonsoft.Json.Schema.yaml
+++ b/curations/nuget/nuget/-/Newtonsoft.Json.Schema.yaml
@@ -5,7 +5,7 @@ coordinates:
 revisions:
   1.0.11:
     licensed:
-      declared: AGPL-3.0-only
+      declared: AGPL-3.0-only OR OTHER
   2.0.10:
     licensed:
       declared: AGPL-3.0-only OR OTHER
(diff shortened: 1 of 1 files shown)


capfei:
@TRQ3 This is also available under a commercial license per the license text:

```
## Commercial License

In addition to this license, Json.NET Schema is offered under a commercial license.
```

Do you agree that this should be `AGPL-3.0-only OR OTHER`?

TRQ3:
@capfei I left as AGPL-3.0 only because "commercial license" is so broad and unclear. Is there a standard approach to the use of that term? If it is to curate as "other," then go for it.

capfei:
`OTHER` is used for when there is a license but does not have SPDX identifier, such as proprietary licenses. When to use `OTHER` is stated in the curation guidelines: https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md

https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md#license-information
https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md#troubleshooting

I hope this helps!

PR #4860, opened 2020-07-09
regex	2020.5.14

mihouc:
**Type:** Missing

**Summary:**
regex	2020.5.14

**Details:**
PyPi site indicates license is Python Software Foundation License

**Resolution:**
Setup.py file also indicates license is Python Software Foundation License

**Affected definitions**:
- [regex 2020.5.14](https://clearlydefined.io/definitions/pypi/pypi/-/regex/2020.5.14/2020.5.14)

### DIFF
--- a/curations/pypi/pypi/-/regex.yaml
+++ b/curations/pypi/pypi/-/regex.yaml
@@ -15,6 +15,9 @@ revisions:
   2020.4.4:
     licensed:
       declared: Python-2.0
+  2020.5.14:
+    licensed:
+      declared: PSF-2.0
   2020.6.8:
     licensed:
       declared: Python-2.0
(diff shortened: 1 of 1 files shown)


capfei:
I don't know if `PSF-2.0` or `Python-2.0` is correct in this instance and know idea if it even makes a difference. I reached out to the community for guidance as I don't have much knowledge regarding the Python license.

curations/pypi/pypi/-/regex.yaml
```diff
@@ -15,6 +15,9 @@ revisions:
   2020.4.4:
     licensed:
       declared: Python-2.0
+  2020.5.14:
+    licensed:
+      declared: PSF-2.0
```
pombredanne, line 20: PSF-2.0 is a rarer thing and is a subset of the complete Python-2.0 license used solo.
Here regex is based on, includes and is derived from the main re Python code and is therefore under the same terms, so Python-2.0 and not the bare PSF

PR #4883, opened 2020-07-09
Prism.Wpf/6.3.0

TRQ3:
**Type:** Missing

**Summary:**
Prism.Wpf/6.3.0

**Details:**
No info in package files. Meta data points to MIT. 

**Resolution:**
https://github.com/PrismLibrary/Prism/blob/master/LICENSE

**Affected definitions**:
- [Prism.Wpf 6.3.0](https://clearlydefined.io/definitions/nuget/nuget/-/Prism.Wpf/6.3.0/6.3.0)

### DIFF
--- a/curations/nuget/nuget/-/Prism.Wpf.yaml
+++ b/curations/nuget/nuget/-/Prism.Wpf.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   6.3.0:
     licensed:
-      declared: MIT
+      declared: Apache-2.0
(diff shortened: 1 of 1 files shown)


capfei:
@TRQ3 It looks like this version was released under Apache 2.0. See https://github.com/PrismLibrary/Prism/blob/Prism.v6.3.0/LICENSE.

Let me know if you agree that it should be curated as Apache 2.0.

TRQ3:
@capfei I agree. Thanks!

PR #4902, opened 2020-07-09
DiffPlex/1.2.1

TRQ3:
**Type:** Missing

**Summary:**
DiffPlex/1.2.1

**Details:**
No info in package files. Meta data shows package moved, but does confirm Apache 2.0 in new location. 

**Resolution:**
https://github.com/mmanela/diffplex/blob/master/License.txt

**Affected definitions**:
- [DiffPlex 1.2.1](https://clearlydefined.io/definitions/nuget/nuget/-/DiffPlex/1.2.1/1.2.1)

### DIFF
--- a/curations/nuget/nuget/-/DiffPlex.yaml
+++ b/curations/nuget/nuget/-/DiffPlex.yaml
@@ -3,6 +3,9 @@ coordinates:
   provider: nuget
   type: nuget
 revisions:
+  1.2.1:
+    licensed:
+      declared: Apache-2.0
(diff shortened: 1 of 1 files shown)


capfei:
@TRQ3 It looks like this version was release under `MS-PL`. See https://github.com/mmanela/diffplex/blob/v1.2.1/License.txt.

Do you agree that this should be curated as `MS-PL`?

TRQ3:
@capfei I do. Apologies. Thank you!

PR #5020, opened 2020-07-15
VSSDK.IDE.12 12.0.4

AE49:
**Type:** Incomplete

**Summary:**
VSSDK.IDE.12 12.0.4

**Details:**
Declaring None

**Resolution:**
Nuspec file in package:

This is a metadata-only package used as a dependency for reference packages and extensions that target Visual Studio 2013 and newer. 

**Affected definitions**:
- [VSSDK.IDE.12 12.0.4](https://clearlydefined.io/definitions/nuget/nuget/-/VSSDK.IDE.12/12.0.4/12.0.4)

### DIFF
--- a/curations/nuget/nuget/-/VSSDK.IDE.12.yaml
+++ b/curations/nuget/nuget/-/VSSDK.IDE.12.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   12.0.4:
     licensed:
-      declared: OTHER
+      declared: NONE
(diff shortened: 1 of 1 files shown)


capfei:
@AE49 I'm not seeing license or source information. Can you point me to it? If there isn't any, should we update the PR to NONE?

AE49:
@capfei I updated to NONE (probably the better option given no actual license file although it may be covered under VS 2013 and newer).

capfei:
@AE49 Thank you for the update!

PR #5026, opened 2020-07-15
Rx-Core/2.2.2

TRQ3:
**Type:** Missing

**Summary:**
Rx-Core/2.2.2

**Details:**
No info in package files. Meta data seems to point to MS EULA, but the Source Repo is MIT. 

**Resolution:**
https://github.com/dotnet/reactive/blob/master/LICENSE

**Affected definitions**:
- [Rx-Core 2.2.2](https://clearlydefined.io/definitions/nuget/nuget/-/Rx-Core/2.2.2/2.2.2)

### DIFF
--- a/curations/nuget/nuget/-/Rx-Core.yaml
+++ b/curations/nuget/nuget/-/Rx-Core.yaml
@@ -6,6 +6,9 @@ revisions:
   2.1.30214:
     licensed:
       declared: Apache-2.0
+  2.2.2:
+    licensed:
+      declared: MIT
   2.2.4:
     licensed:
       declared: Apache-2.0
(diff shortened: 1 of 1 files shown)


capfei:
@TRQ3 There isn't a tag for this version, but looking at the repo around the time this version was published, it looks like the license was Apache 2.0.

https://github.com/dotnet/reactive/blob/044afd56c6e3c6dc30ee8baefb45ef4a43ce47d1/Rx.NET/Source/license.txt

Do you agree this should be curated as Apache instead of MIT?

TRQ3:
@capfei That looks right to me. Thanks.

PR #5035, opened 2020-07-15
loadimpact/k6479707f90eae5

ariel11:
**Type:** Missing

**Summary:**
loadimpact/k6479707f90eae5

**Details:**
https://github.com/loadimpact/k6/blob/479707f90eae51ee385b3424e6a1259702608e8a/LICENSE.md

**Resolution:**
AGPL-3.0-only

**Affected definitions**:
- [k6 479707f90eae51ee385b3424e6a1259702608e8a](https://clearlydefined.io/definitions/git/github/loadimpact/k6/479707f90eae51ee385b3424e6a1259702608e8a/479707f90eae51ee385b3424e6a1259702608e8a)

### DIFF
--- a/curations/git/github/loadimpact/k6.yaml
+++ b/curations/git/github/loadimpact/k6.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   479707f90eae51ee385b3424e6a1259702608e8a:
     licensed:
-      declared: AGPL-3.0-only
+      declared: AGPL-3.0-or-later
--- a/curations/git/github/loadimpact/k6.yaml
+++ b/curations/git/github/loadimpact/k6.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: k6
+  namespace: loadimpact
+  provider: github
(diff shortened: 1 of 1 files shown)


capfei:
@ariel11 Since the file headers stated AGPL-3.0-or-later, should we curate it as such? As far as I can tell, the file header is what differentiates them since the license text appears to be the same. Please correct me if I'm wrong!

https://spdx.org/licenses/AGPL-3.0-only.html and https://spdx.org/licenses/AGPL-3.0-or-later.html both state

> The license notice (as seen in the Standard License Header field below) states which of these applies to the code in the file.

Randomly selected files:
https://github.com/loadimpact/k6/blob/479707f90eae51ee385b3424e6a1259702608e8a/main.go
https://github.com/loadimpact/k6/blob/479707f90eae51ee385b3424e6a1259702608e8a/core/local/local.go

ariel11:
@capfei - Yes, I agree with you.  This should be "or later" based on the file headers.  Thank you!

PR #5085, opened 2020-07-20
distlib/0.3.1

TRQ3:
**Type:** Missing

**Summary:**
distlib/0.3.1

**Details:**
No info in package files. Meta data points to Python Software Foundation License 2.0, but found source repo confirms. The "Declared" field would not accept the SPDX abbreviation "PSF-2.0"

**Resolution:**
https://github.com/vsajip/distlib/blob/0.3.1/LICENSE.txt

**Affected definitions**:
- [distlib 0.3.1](https://clearlydefined.io/definitions/pypi/pypi/-/distlib/0.3.1/0.3.1)

### DIFF
--- a/curations/pypi/pypi/-/distlib.yaml
+++ b/curations/pypi/pypi/-/distlib.yaml
@@ -6,3 +6,6 @@ revisions:
   0.3.0:
     licensed:
       declared: Python-2.0
+  0.3.1:
+    licensed:
+      declared: '[object Object]'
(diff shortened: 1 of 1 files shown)


capfei:
@TRQ3 The identifier `PSF-2.0` is for projects that only use a small portion of the Python 2.0 license. See https://spdx.org/licenses/PSF-2.0.html.

When the full Python 2.0 license is used, like it is for this component, curate it is as `Python-2.0`.
https://spdx.org/licenses/Python-2.0.html

capfei:
PyPi doesn't have separate, supported license identifiers for Python 2.0 and PSF 2.0. They both get listed as `License :: OSI Approved :: Python Software Foundation License`.

From my understanding, they typically mean Python 2.0 as PSF 2.0 is rarely used.

PR #5101, opened 2020-07-20
Declare MIT

peggymoloney:
**Type:** Missing

**Summary:**
Declare MIT

**Details:**
Declare MIT found here (https://github.com/Azure/azure-functions-vs-build-sdk/blob/main/LICENSE)

**Resolution:**
Specific tag not identified but no change to the license since the beginning of the project 

**Affected definitions**:
- [Microsoft.NET.Sdk.Functions 1.0.2](https://clearlydefined.io/definitions/nuget/nuget/-/Microsoft.NET.Sdk.Functions/1.0.2/1.0.2)

### DIFF
--- a/curations/nuget/nuget/-/Microsoft.NET.Sdk.Functions.yaml
+++ b/curations/nuget/nuget/-/Microsoft.NET.Sdk.Functions.yaml
@@ -12,6 +12,9 @@ revisions:
   1.0.19:
     licensed:
       declared: MIT
+  1.0.2:
+    licensed:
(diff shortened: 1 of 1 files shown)


ariel11:
It looks like this package version pre-dates the license in the repo, however - per #3 of the curation guidelines (copied below), going to merge this as MIT, per later versions.

#3 - Target component has no project/source location and does not have any license information. There exists a related version (newer or older) that does have license information.
Use the related component’s license data, favoring for newer version if possible to discern

PR #5121, opened 2020-07-20
net.sf.py4j:py4j	0.10.4

mihouc:
**Type:** Missing

**Summary:**
net.sf.py4j:py4j	0.10.4

**Details:**
.pom file indicates license is BSD-2

**Resolution:**
 

**Affected definitions**:
- [py4j 0.10.4](https://clearlydefined.io/definitions/maven/mavencentral/net.sf.py4j/py4j/0.10.4/0.10.4)

### DIFF
--- a/curations/maven/mavencentral/net.sf.py4j/py4j.yaml
+++ b/curations/maven/mavencentral/net.sf.py4j/py4j.yaml
@@ -6,7 +6,7 @@ coordinates:
 revisions:
   0.10.4:
     licensed:
-      declared: BSD-2-Clause
+      declared: BSD-3-Clause
   0.10.6:
     licensed:
       declared: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


capfei:
@mihouc This looks to be BSD-3-Clause instead of BSD-2-Clause. The files in the source have the BSD-3-Clause license in their headers. The name they use in the pom, The New BSD License, is also BSD-3-Clause. 

The link they use did point to BSD-3-Clause at some point but I don't know when it was changed to point to BSD-2-Clause.

https://github.com/bartdag/py4j/blob/0.10.4/py4j-java/LICENSE.txt
https://github.com/bartdag/py4j/blob/0.10.4/py4j-java/src/main/java/py4j/Base64.java

Do you agree this should be BSD-3-Clause?

PR #5185, opened 2020-07-22
regex 2020.5.7

ariel11:
**Type:** Missing

**Summary:**
regex 2020.5.7

**Details:**
The metadata says Python Software Foundation License, but the py file headers say: # This version of the SRE library can be redistributed under CNRI's
# Python 1.6 license.  For any other use, please contact Secret Labs
# AB (info@pythonware.com).

**Resolution:**
Based on the above, selecting CNRI-Python

**Affected definitions**:
- [regex 2020.5.7](https://clearlydefined.io/definitions/pypi/pypi/-/regex/2020.5.7/2020.5.7)

### DIFF
--- a/curations/pypi/pypi/-/regex.yaml
+++ b/curations/pypi/pypi/-/regex.yaml
@@ -18,6 +18,9 @@ revisions:
   2020.5.14:
     licensed:
       declared: Python-2.0
+  2020.5.7:
+    licensed:
+      declared: CNRI-Python
   2020.6.8:
     licensed:
       declared: Python-2.0


capfei:
@ariel11 From a previous discussion about regex, it was concluded that it should be Python 2.0.

> Here regex is based on, includes and is derived from the main re Python code and is therefore under the same terms, so Python-2.0

https://github.com/clearlydefined/curated-data/pull/4860

Let me know if you agree.

ariel11:
Well, the top of the regex_core py file says this.  By "CNRI's Python 1.6 license," I thought they meant this  - https://spdx.org/licenses/CNRI-Python.html#licenseText
image

Thoughts?

capfei:
😕  I wish I knew more about the history of Python. 

According to the author, he based it off of the re module in Python, which uses the Python license.
https://bitbucket.org/mrabarnett/mrab-regex/issues/362/any-license-work-for-this-project

However, if I'm looking at the right [file in Python](https://github.com/python/cpython/blob/master/Lib/re.py), it has the same CNRI header. But, then I read https://www.python.org/download/releases/1.6.1/, which states:

> The new regular expression engine, SRE by Fredrik Lundh, is fully backwards compatible with the old engine, and is in fact invoked using the same interface (the "re" module). You can explicitly invoke the old engine by import pre, or the SRE engine by importing sre. SRE is faster than pre, and supports Unicode (which was the main reason to put effort in yet another new regular expression engine -- this is at least the fourth!).

So I'm not sure if SRE is re or if SRE is built on re... 😖

capfei:
If other would like to add their thoughts, it will be greatly appreciated!

PR #5838, opened 2020-08-31
bcprov-jdk15on 1.65.01

msohn:
Add
- declare license
- source location
- urls

### DIFF
--- a/curations/maven/mavencentral/org.bouncycastle/bcprov-jdk15on.yaml
+++ b/curations/maven/mavencentral/org.bouncycastle/bcprov-jdk15on.yaml
@@ -25,8 +25,6 @@ revisions:
     licensed:
       declared: MIT
   '1.65.01':
-    licensed:
-      declared: MIT
     sourceLocation:
       name: bcprov-jdk15on
       namespace: org.bouncycastle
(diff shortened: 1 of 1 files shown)


capfei:
FYI: I'm having someone take a look at the definition and figure out why it is giving an error.

curations/maven/mavencentral/org.bouncycastle/bcprov-jdk15on.yaml
```diff
@@ -24,6 +24,22 @@ revisions:
...
+      declared: MIT
+    sourceLocation:
+      name: bcprov-jdk15on
+      namespace: org.bouncycastle
+      provider: mavencentral
+      revision: '1.65.01'
+      type: sourcearchive
+      url: >-
+        https://search.maven.org/remotecontent?filepath=org/bouncycastle/bcprov-jdk15on/1.65.01/bcprov-jdk15on-1.65.01-sources.jar
+    urls:
```
capfei, line 38: It looks like this part is not curate-able. Can I delete this so I can merge the rest?

msohn: sure, the documentation what exactly can be curated and which fields are added by some bot isn't very clear and the UI doesn't allow to specify a source artefact from maven central but only a git repository

capfei (approved):

PR #6030, opened 2020-09-09
There is no license specified on the file

waynebeaton:
**Type:** Incorrect

**Summary:**
There is no license specified on the file

**Details:**
This file describes the history of the content. It does have entries that discuss licenses, but does not itself have any header with any indication of how the file is licensed.

**Resolution:**
Change the license for the single file to NONE.

**Affected definitions**:
- [file-uri-to-path 1.0.0](https://clearlydefined.io/definitions/npm/npmjs/-/file-uri-to-path/1.0.0/1.0.0)

### DIFF
--- a/curations/npm/npmjs/-/file-uri-to-path.yaml
+++ b/curations/npm/npmjs/-/file-uri-to-path.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: file-uri-to-path
+  provider: npmjs
+  type: npm
+revisions:
+  1.0.0:
+    files:
+      - license: NONE
+        path: package/History.md


capfei:
@waynebeaton Wouldn't the MIT license apply to this file? Most of the files do not have file headers but, I would assume the license file included with the package applies.

waynebeaton:
Yes. I believe that we can reasonably assert that the MIT license applies.

But the file itself does not specify this. My understanding is that the intention of the field that I've modified is to capture the license of the corresponding file itself. The file does not express a license so this field should be "NONE". Please correct me if I'm wrong.

capfei:
@waynebeaton I think you are correct. I got too focused on the license file since that is the most of the PRs. 😞

capfei (approved):

waynebeaton:
@capfei, I'm super excited to learn that I understand something :-)

PR #6044, opened 2020-09-09
LibGit2Sharp.NativeBinaries	1.0.185

mihouc:
**Type:** Missing

**Summary:**
LibGit2Sharp.NativeBinaries	1.0.185

**Details:**
License link from Nuget site indicates license is GPL-2.0-with-GCC-exception

**Resolution:**
Link to copying file in Nuspec file also indicates license is GPL-2.0-with-GCC-exception

**Affected definitions**:
- [LibGit2Sharp.NativeBinaries 1.0.185](https://clearlydefined.io/definitions/nuget/nuget/-/LibGit2Sharp.NativeBinaries/1.0.185/1.0.185)

### DIFF
--- a/curations/nuget/nuget/-/LibGit2Sharp.NativeBinaries.yaml
+++ b/curations/nuget/nuget/-/LibGit2Sharp.NativeBinaries.yaml
@@ -8,7 +8,7 @@ revisions:
       declared: GPL-2.0-only WITH OTHER
   1.0.185:
     licensed:
-      declared: GPL-2.0-with-GCC-exception
+      declared: GPL-2.0-only WITH OTHER
   1.0.81:
     licensed:
       declared: GPL-2.0-only WITH OTHER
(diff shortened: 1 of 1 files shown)


capfei:
@mihouc `GPL-2.0-with-GCC-exception` is a deprecated license identifier and shouldn't be used. See https://spdx.org/licenses/GPL-2.0-with-GCC-exception.html.

This should be curated as `GPL-2.0-only AND OTHER` since the linking exception text is slightly different than the [GCC exception 2.0 text](https://spdx.org/licenses/GCC-exception-2.0.html).

LibGit2
> In addition to the permissions in the GNU General Public License, the _**authors give**_ you unlimited permission to link the compiled version of this _**library**_


GCC Linking Exception 2.0
> In addition to the permissions in the GNU General Public License, the _**Free Software Foundation gives**_ you unlimited permission to link the compiled version of this _**file**_

If you agree, I will update the PR.

mihouc:
I agree with the revision, thanks for catching the error

capfei (approved):

PR #6463, opened 2020-09-22
Bogus/29.0.2

TRQ3:
**Type:** Missing

**Summary:**
Bogus/29.0.2

**Details:**
Meta data points to both MIT and BSD 3 Clause. 

**Resolution:**
https://raw.githubusercontent.com/bchavez/Bogus/master/LICENSE

**Affected definitions**:
- [Bogus 29.0.2](https://clearlydefined.io/definitions/nuget/nuget/-/Bogus/29.0.2/29.0.2)

### DIFF
--- a/curations/nuget/nuget/-/Bogus.yaml
+++ b/curations/nuget/nuget/-/Bogus.yaml
@@ -6,6 +6,9 @@ revisions:
   22.0.8:
     licensed:
       declared: MIT
+  29.0.2:
+    licensed:
+      declared: MIT AND BSD-3-Clause
   30.0.2:
     licensed:
       declared: MIT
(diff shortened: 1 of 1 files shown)


AE49:
@TRQ3 @capfei I think we would only Declare MIT (bottom of this page only MIT is listed -https://github.com/bchavez/Bogus). BSD-3-Clause would be a discovered license. Do you agree?

capfei:
@AE49 Yes, I agree that it should MIT only since it looks like the BSD-3-Clause license info applies to https://github.com/pid/speakingurl/tree/563da7ebd8c957804de43a92359d6e53796c3226 in https://github.com/bchavez/Bogus/tree/v29.0.2/Source.

AE49 (approved):

PR #7269, opened 2020-10-08
Update source link

waynebeaton:
**Type:** Incorrect

**Summary:**
Update source link

**Details:**
The source link is not valid.

**Resolution:**
Set the source link to point to GitHub repository.

**Affected definitions**:
- [cxf-rt-frontend-jaxrs 3.3.4](https://clearlydefined.io/definitions/maven/mavencentral/org.apache.cxf/cxf-rt-frontend-jaxrs/3.3.4/3.3.4)

### DIFF
--- a/curations/maven/mavencentral/org.apache.cxf/cxf-rt-frontend-jaxrs.yaml
+++ b/curations/maven/mavencentral/org.apache.cxf/cxf-rt-frontend-jaxrs.yaml
@@ -1,0 +1,18 @@
+coordinates:
+  name: cxf-rt-frontend-jaxrs
+  namespace: org.apache.cxf
+  provider: mavencentral
+  type: maven
+revisions:
+  3.3.4:
+    described:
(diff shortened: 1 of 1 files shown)


waynebeaton:
The source repository has a lot more content than just what's in this JAR (the source that this JAR actually cares about is in cxf/rt/frontend/jaxrs/src/main. I thought that we might be able to use facets to narrow this down, but it's not clear to me how we do that.

The DEPENDENCIES file, which shows up as being multiply licensed, lists licenses but has no license declaration of its own, so I've marked it as "NONE". The file doesn't appear to be in the Git repository (it's may created as part of the build or something). The sources JAR is on Maven Central.

https://repo1.maven.org/maven2/org/apache/cxf/cxf-rt-frontend-jaxrs/3.3.4/cxf-rt-frontend-jaxrs-3.3.4-sources.jar

capfei (approved):

PR #7431, opened 2020-10-13
Revert "dotnet/runtime 38017c3935de95d0335bac04f4901ddfc2718656"

AE49:
Reverts clearlydefined/curated-data#7179

### DIFF
--- a/curations/git/github/dotnet/runtime.yaml
+++ b/curations/git/github/dotnet/runtime.yaml
@@ -1,9 +1,0 @@
-coordinates:
-  name: runtime
-  namespace: dotnet
-  provider: github
-  type: git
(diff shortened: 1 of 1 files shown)


AE49:
Should add OTHER for the Patent.txt

AE49:
@capfei I merged this and realized after there is a Patent.txt file and I need to add OTHER for it. I did a Revert thinking it would give me the option to fix the file and approve again but that doesn't seem to be right. How do I make the fix?

jeffmendoza:
Just make a new curation to add OTHER

capfei:
@AE49 I agree with @jeffmendoza

PR #7659, opened 2020-10-23
File that is listed as MPL-2.0 isn't in the source

waynebeaton:
**Type:** Incorrect

**Summary:**
File that is listed as MPL-2.0 isn't in the source

**Details:**
I pulled the listed source JAR and the file that is listed as MPL-2.0 isn't in the source, AFAICT. I don't see the file in the [GitHub source](https://github.com/apache/httpcomponents-client/tree/3734aaa038a58c17af638e9fa29019cacb22e82c) either.

**Resolution:**
Mark the file as NONE. A better solution would be to remove it.

**Affected definitions**:
- [httpclient5 5.0.3](https://clearlydefined.io/definitions/maven/mavencentral/org.apache.httpcomponents.client5/httpclient5/5.0.3/5.0.3)

### DIFF
--- a/curations/maven/mavencentral/org.apache.httpcomponents.client5/httpclient5.yaml
+++ b/curations/maven/mavencentral/org.apache.httpcomponents.client5/httpclient5.yaml
@@ -1,0 +1,10 @@
+coordinates:
+  name: httpclient5
+  namespace: org.apache.httpcomponents.client5
+  provider: mavencentral
+  type: maven
+revisions:
+  5.0.3:
+    files:
+      - license: NONE
+        path: mozilla/public-suffix-list.txt


ariel11:
@waynebeaton - thank you.  I don't know where the tooling is picking up the public-suffix-list.txt.  I will open an issue for the ClearlyDefined engineers.  

FYI:  @nellshamrell

ariel11:
Opened https://github.com/clearlydefined/service/issues/772

ariel11:
@nellshamrell  - do you have any info on this?  waynebeaton reports the tooling is _erroneously_ including the public-suffix-list.txt file in the package contents under "Files," and that this file is not actually in this package?  Can you look into, so I can either close this PR or we can see what needs to be fixed?  Thank you!

nellshamrell:
I haven't had a chance to look into this yet, but I just assigned myself to [the issue](https://github.com/clearlydefined/service/issues/772)

waynebeaton:
> @nellshamrell - do you have any info on this? waynebeaton reports the tooling is _erroneously_ including the public-suffix-list.txt file in the package contents under "Files," and that this file is not actually in this package? Can you look into, so I can either close this PR or we can see what needs to be fixed? Thank you!

FWIW, I said that the file isn't in the source. 

I took a harder look at the source, and it appears as though the file is pulled in as part of the build. So it's not in the source, but it is in the distribution JAR.

I'm thinking that this curation is bogus and that I should withdraw it.

ariel11:
@waynebeaton - got it.  I will close this PR.  Thanks for taking another look!

@nellshamrell - FYI

Issue #7836, opened 2020-10-30
Disallow the use of "OTHER" as a declared license


sschuberth:
Several curations use "OTHER" as the declared license, e.g. https://github.com/clearlydefined/curated-data/blob/da3aff50f7b35dca51116550f671e84255203451/curations/npm/npmjs/-/jsonify.yaml#L8

First of all, there is a general problem as "OTHER" is not a valid SPDX expression. Secondly, at the concrete example of `jsonify`, consuming the ClearlyDefined curation worsens the meta data from "Public Domain" as declared in its [package.json](https://github.com/substack/jsonify/blob/7064ab53e5f73fa31e2ce7aa47d210c539662f16/package.json#L29) to "OTHER", which is even less telling, and causes [ORT](https://github.com/oss-review-toolkit/ort) (which has a [mapping](https://github.com/oss-review-toolkit/ort/blob/9c9ecc0c7e7dd6e39c7b09cf02616353b5269eb3/spdx-utils/src/main/resources/declared-license-mapping.yml#L351) from "Public Domain" to "LicenseRef-scancode-public-domain-disclaimer") to run into issues.

That's why I'd like to prose to not use "OTHER" at all. What do you think @capfei @fossygirl?

fossygirl:
@sschuberth I absolutely see your points. This was a request from our curation community. I'll let @ariel11 weigh in.

ariel11:
@sschuberth and @fossygirl - ClearlyDefined uses several non-SPDX identifiers:  OTHER, NONE, and NOASSERTION.  These are further explained in the curation guidelines [here](https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md#license-information). 

The challenge we have is a fair number of projects/components do not have a license with a SPDX match.  When there is license information but the tooling is not able to tell what the license is, it says NOASSERTION.  We want a way to be able to say "hey, a human looked at this, there is license text/language here, but there's not a SPDX for it."  That's when we use "OTHER."  

I agree, "OTHER" becomes a large bucket that can cover multiple scenarios  - it could be for projects that say they are in the public domain, or projects with proprietary or commercial licenses, or projects that modified their license text so it no longer fits the SPDX matching guidelines. 

[...]

PR #7937, opened 2020-11-05
Add missing dashes to SPDX expressions in various Maven curations

sschuberth:
FYI @nnobelis.

### DIFF
--- a/curations/maven/mavencentral/com.sun.codemodel/codemodel.yaml
+++ b/curations/maven/mavencentral/com.sun.codemodel/codemodel.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   '2.6':
     licensed:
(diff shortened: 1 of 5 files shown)


sschuberth:
@jeffmcaffer isn't there a way we could have a PR check to validate that curated license expressions are actually valid SPDX expressions (modulo maybe `OTHER`, also see [our discussion](https://github.com/clearlydefined/curated-data/issues/7836) in that regard)?

capfei (approved):

capfei:
I would love to have expression validation! I have made my fair share of typos when making corrections. 😞

jeffmcaffer:
PR validation is a great idea. we have an spdx library that should be able to do that work so putting together an action should be straight forward. 

cc @nellshamrell

PR #8188, opened 2020-11-20
Fixing psycopg2-binary 2.8.4

ariel11:
**Type:** Incorrect

**Summary:**
Fixing psycopg2-binary 2.8.4

**Details:**
Declared is incorrectly GPL when it should be LGPL.  See LICENSE file in "Files" section. 

**Resolution:**
Curated:  LGPL-3.0-or-later  (Note:  a BSD license applies to some files, but I think that would be considered a "Discovered" license).  Thoughts?

**Affected definitions**:
- [psycopg2-binary 2.8.4](https://clearlydefined.io/definitions/pypi/pypi/-/psycopg2-binary/2.8.4/2.8.4)

### DIFF
--- a/curations/pypi/pypi/-/psycopg2-binary.yaml
+++ b/curations/pypi/pypi/-/psycopg2-binary.yaml
@@ -5,7 +5,7 @@ coordinates:
 revisions:
   2.8.4:
     licensed:
-      declared: LGPL-3.0-or-later
+      declared: LGPL-3.0-or-later WITH OTHER
   2.8.5:
     licensed:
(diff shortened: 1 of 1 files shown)


capfei:
@ariel11 Should we add `WITH OTHER` since it looks like they included an exception:

```
In addition, as a special exception, the copyright holders give
permission to link this program with the OpenSSL library (or with
modified versions of OpenSSL that use the same license as OpenSSL),
and distribute linked combinations including the two.
```

ariel11:
@capfei - yes, I agree adding "WITH OTHER."  Thanks

capfei (approved):

PR #9241, opened 2021-01-22
License is MPL-1.1

waynebeaton:
**Type:** Incorrect

**Summary:**
License is MPL-1.1

**Details:**
I believe that the "MPL" content cited in the header is from here: https://gist.github.com/xgz123/3c039d60bdeab36bc9082f134c2d1953 (which is licensed MPL-1.1)

**Resolution:**
Change the license on this one file to MPL-1.1.

**Affected definitions**:
- [html-minifier 3.5.21](https://clearlydefined.io/definitions/npm/npmjs/-/html-minifier/3.5.21/3.5.21)

### DIFF
--- a/curations/npm/npmjs/-/html-minifier.yaml
+++ b/curations/npm/npmjs/-/html-minifier.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: html-minifier
+  provider: npmjs
+  type: npm
+revisions:
+  3.5.21:
+    files:
+      - license: MPL-1.1
+        path: package/src/htmlparser.js


capfei:
I agree with MPL-1.1. Providing how I came to this in case anyone is curious:

It looks like when this was originally published by Erik back in 2004, it was only under MPL 1.1. According to the Wayback Machine, it wasn't changed to dual license until 2010 and then triple licensed in 2011.

Original post:
https://web.archive.org/web/20041127223447/http://erik.eae.net/

So I can only assume that when John Resig modified it in 2008, it was still only under MPL 1.1.
https://johnresig.com/blog/pure-javascript-html-parser/

capfei (approved):

PR #10486, opened 2021-04-05
matplotlib 3.4.0

clearlydefinedbot:
**Type:** Missing

**Summary:**
matplotlib 3.4.0

**Details:**
Add 0BSD License

**Resolution:**
License Url: 
https://matplotlib.org/stable/devel/license.html

Description:

Pull request generated by Microsoft tooling.

**Affected definitions**:
- [matplotlib 3.4.0](https://clearlydefined.io/definitions/pypi/pypi/-/matplotlib/3.4.0)


AE49:
@capfei This could be Declared as PSF-2.0. PyPI metadata is PSF. The license is a close enough match with references to PSF and Python customized, but main licensing text was not changed. https://matplotlib.org/stable/users/license.html. Thoughts?

https://github.com/matplotlib/matplotlib/blob/v3.4.0/LICENSE/LICENSE

capfei:
@ariel11 I do see that only the author information was changes and the copyright years, which, in most cases, we would accept as being that license. I'm not sure about the part referring versions prior to 1.3.0. Do we count that as part of the full license text or treat it individually?

@ariel11 What do you think?

ariel11:
@AE49 and @capfei - this might be a good one for the Discord call.  On the one hand, the Matplotlib license does not meet the SPDX matching guidelines as the author text is not in colored font - https://spdx.org/licenses/PSF-2.0.html.  I think it _should_ be in colored font and I can submit that request to SPDX.  It looks like we curated a different version of Matplotlib as OTHER.

AE49:
@ariel11 @capfei Are we waiting first on the SPDX resolution or should I merge this PR now with PSF-2.0 as the Declared License?

ariel11:
@AE49 - My suggestion would be to proceed with PSF-2.0

AE49 (approved):

capfei:
I tried to fix these but it keeps failing. Added to #11048

ariel11 (approved):

PR #12353, opened 2021-05-24
Updating WITH OTHER

capfei:


ariel11 (approved):

ariel11:
@capfei  - These look right to me - a project with a PATENTS file would use "AND OTHER."  And ClearlyDefined, as a project, agreed stop curating "WITH OTHER" for licenses with exception text that does not have a valid SPDX exception ID (so they now become "OTHER:).

PR #12525, opened 2021-06-06
Defined new revisions for golang dep: github.com/google/shlex

AndrienkoAleksandr:


capfei:
Ran component through the scanners since I noticed it had not been harvested yet. PR matches curation site. Closing.

PR #12611, opened 2021-06-06
Defined new revisions for golang dep: github.com/otiai10/copy

AndrienkoAleksandr:


capfei:
Ran component through the scanners since I noticed it had not been harvested yet and it correctly declared the license. Closing PR as it is no longer needed.

PR #12622, opened 2021-06-06
Defined new revisions for golang dep: github.com/smartystreets/assertions

AndrienkoAleksandr:


capfei:
@ariel11 The last part in the license reads more informational instead of adding additional license verbiage. What do you think? I think this should just be MIT.

ariel11:
@capfei - Yes I agree that for SPDX matching purposes, the NOTE would be "extraneous text" per the SPDX matching guidelines and we would just curate "MIT" here as the declared license. 

image

capfei (approved):

PR #12624, opened 2021-06-06
Defined new revisions for golang dep: github.com/storageos/go-api

AndrienkoAleksandr:


ariel11:
@AndrienkoAleksandr - I think with the way ClearlyDefined defines "declared" vs. "discovered" licenses, MIT may be the "declared" license for this StorageOS project and it includes code from go-dockerclient, which is BSD-2-Clause.  

Thoughts @capfei ?

ariel11 (approved):

PR #13378, opened 2021-07-13
Update the source information

msftenhanceprovenance:
**Type:** Missing

**Summary:**
Update the source information

**Details:**
With the rise in supply chain attacks and OSS dependencies being used as a attack vector, Microsoft is working with our ecosystem partners, such as the Linux Foundation's OpenSSF, to enable OSS consumers to track packages back to their public sources.

**Resolution:**
This PR was created to add this value, ensuring future releases will include this provenance information.
Published NPM packages with repository information:
	* has-binary

**Affected definitions**:
- [has-binary 0.1.5](https://clearlydefined.io/definitions/npm/npmjs/-/has-binary/0.1.5)

### DIFF
--- a/curations/npm/npmjs/-/has-binary.yaml
+++ b/curations/npm/npmjs/-/has-binary.yaml
@@ -1,0 +1,32 @@
+coordinates:
+  name: has-binary
+  provider: npmjs
+  type: npm
+revisions:
+  0.1.5:
+    described:
+      sourceLocation:
+        name: has-binary
+        namespace: socketio
+        provider: github
+        revision: 9034d3f6822521245b2507633b7d4eb54dc5ec91
+        type: git
+        url: 'https://github.com/socketio/has-binary/commit/9034d3f6822521245b2507633b7d4eb54dc5ec91'
+  0.1.6:
+    described:
+      sourceLocation:
+        name: has-binary
+        namespace: socketio
+        provider: github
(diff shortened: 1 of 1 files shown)


capfei:
@ariel11 When deciding the source location, we should go with where the code currently lives, right? In this case, the latest repo is a fork of a fork. socketio/has-binary is a fork of Automattic/has-binary, which is a fork of kevin-roark/has-binarydata.js.

- Source repo (kevin-roark/has-binarydata.js): Package.json lists a different name so not what I would consider this as the source for has-binary NPM package.
- Fork #1 (Automattic/has-binary): This looks like the repo they were working out of for version 0.1.5 and 0.1.6 based on the creation date of this fork (Sept 4, 2014).
- Fork #2 (socketio/has-binary): This is a fork of Automattic/has-binary and where version 0.1.7 is published from. This was created on Nov 18, 2015.

What do you think?

ariel11:
@capfei - I think we'd want to go with the repo used at the time the package was published, so the real/actual source location that was used to build the package.  Are we able to tell that?  If not, or its not clear, let's circle back with the team we met with and ask for their feedback.  Sound good?

ariel11:
@capfei - I just sent an email to the folks behind this PR for their thoughts/guidance on your question and copied you.  Thank you

msftenhanceprovenance:
Update SrcRepo and url.

PR #13582, opened 2021-07-21
Discovered license is incorrect

sophokles73:
**Type:** Incorrect

**Summary:**
Discovered license is incorrect

**Details:**
The GPL-2.0-only license does not apply here because the dependency that it comes from (jboss-transaction-api_1.2_spec) is licensed under either CDDL or GPL 2.0 as can be seen here: https://github.com/jboss/jboss-transaction-api_spec/tree/jboss-transaction-api_1.2_spec-1.1.1.Final

**Resolution:**
Set declared and discovered license to Apache-2.0 AND CDDL-1.1

**Affected definitions**:
- [infinispan-client-hotrod 12.0.2.Final](https://clearlydefined.io/definitions/maven/mavencentral/org.infinispan/infinispan-client-hotrod/12.0.2.Final/12.0.2.Final)

### DIFF
--- a/curations/maven/mavencentral/org.infinispan/infinispan-client-hotrod.yaml
+++ b/curations/maven/mavencentral/org.infinispan/infinispan-client-hotrod.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: infinispan-client-hotrod
+  namespace: org.infinispan
+  provider: mavencentral
+  type: maven
+revisions:
+  12.0.2.Final:
+    licensed:
+      declared: Apache-2.0 AND CDDL-1.1


LandOfBliss:
Looked in the "Files" section on the ClearlyDefined definition.  License file indicates Apache-2.0
Pom file on Maven.org does not indicate license
Conclusion:  Declared license is Apache-2.0

sophokles73:
That is true. However, the `META-INF/licenses.xml` file provides an overview of the artifacts dependencies and those include *jboss-transaction-api_1.2_spec* which is dual licensed. That is why the discovered license(s) include GPL-2.0 (one of the available choices for jboss-transaction-api_1.2_spec). My understanding is that infinispan-client-hotrod 12.0.2.Final cannot be licensed under Apache 2.0 if it depends on a GPL-2.0 component. Thus I conclude that the authors of infinispan-client-hotrod have chosen to use jboss-transaction-api_1.2_spec under CDDL. I therefore suggest to update the *discovered license* to Apache 2.0 AND CDDL. WDYT?

ariel11:
@capfei - why am I not seeing the licenses.xml file (in ClearlyDefined) in the https://repo1.maven.org/maven2/org/infinispan/infinispan-client-hotrod/12.0.2.Final/infinispan-client-hotrod-12.0.2.Final-sources.jar?  I'm also not able to open the file in ClearlyDefined.

@sophokles73 - ClearlyDefined's aim is to simply report what the authors say the license is _vs._ doing a legal license on whether the overall package license can be Apache-2.0 if it includes GPL code or not.  I am having trouble finding the licenses.xml file in the sources.jar to see what it says, but if it says the main package license is Apache-2.0 and portions of the package are GPL, the "declared" license ClearlyDefined would report would be Apache-2.0.

ariel11:
Found the licenses.xml file in the .jar and confirmed it seems to only list _dependency_ licenses, which would all be "discovered" licenses.  Do not see that the author stated what the overall "declared" license is for this package.  Closing PR.  Thank you

PR #13590, opened 2021-07-22
Revert "Reprocess merged curation for nuget/nuget/-/protobuf-net/2.0.…

AE49:
…0.668"


AE49:
I may have merged this incorrectly under Apache-2.0. It should be BSD-3-Clause

PR #13779, opened 2021-08-07
 jQuery.vsdoc 1.5.1

LandOfBliss:
**Type:** Missing

**Summary:**
 jQuery.vsdoc 1.5.1

**Details:**
Nuget indicates package may be deprecated.  Nuget does not include source code project or license field links.  No license found in nuspec.

**Resolution:**
Declared license is NONE

**Affected definitions**:
- [jQuery.vsdoc 1.5.1](https://clearlydefined.io/definitions/nuget/nuget/-/jQuery.vsdoc/1.5.1/1.5.1)

### DIFF
--- a/curations/nuget/nuget/-/jQuery.vsdoc.yaml
+++ b/curations/nuget/nuget/-/jQuery.vsdoc.yaml
@@ -3,6 +3,9 @@ coordinates:
   provider: nuget
   type: nuget
 revisions:
+  1.5.1:
+    licensed:
+      declared: NONE
   1.6.0:
     licensed:
       declared: MIT
--- a/curations/nuget/nuget/-/jQuery.vsdoc.yaml
+++ b/curations/nuget/nuget/-/jQuery.vsdoc.yaml
@@ -8,4 +8,4 @@ revisions:
       declared: NONE
   1.6.0:
     licensed:
-      declared: MIT
+      declared: NONE


ariel11:
Hi @LandOfBliss  - if you look in the "Content" folder in ClearlyDefined, it looks like this is MIT: [jquery-1.5.1-vsdoc.js](https://clearlydefined.io/file/c431be7e4134e8d5b7f9b4e701724fb71da09edbe0ec0dc480cdebccc2201f0b).

@capfei  - would you do "MIT" here based on the above?

capfei:
@ariel11 This looks like redistributed third-party code, jQuery and Sizzle.js, with some PowerShell scripts added to add VS 2010 support. Since the owners didn't point to either in the Nuspec, should we leave this as NONE with MIT and BSD left as "Discovered"?

ariel11:
There's also this comment on the files in the "content" and "tools" folder that one could argue for curating this as "OTHER" - 
image

"NONE" also makes sense to me though, as there doesn't seem to be a clear overarching package license.

@royaljust - do you want to weigh in on this one?  

Thanks

royaljust:
I think none :)

ariel11 (approved):

ariel11:
Agree overall package license is NONE and the scripts are 3rd party "discovered" licenses

PR #13841, opened 2021-08-12
com.github.jnr/jnr-posix3.1.4

ariel11:
**Type:** Incorrect

**Summary:**
com.github.jnr/jnr-posix3.1.4

**Details:**
POM lists 3 licenses and in the repo, it is explained this is tri-licensed: https://github.com/jnr/jnr-posix/blob/master/LICENSE.txt

**Resolution:**
see above

**Affected definitions**:
- [jnr-posix 3.1.4](https://clearlydefined.io/definitions/maven/mavencentral/com.github.jnr/jnr-posix/3.1.4/3.1.4)

### DIFF
--- a/curations/maven/mavencentral/com.github.jnr/jnr-posix.yaml
+++ b/curations/maven/mavencentral/com.github.jnr/jnr-posix.yaml
@@ -7,3 +7,6 @@ revisions:
   3.0.45:
     licensed:
       declared: EPL-1.0 OR CPL-1.0 OR GPL-2.0 OR LGPL-2.1
+  3.1.4:
+    licensed:
+      declared: EPL-2.0 OR GPL-2.0-only


capfei:
@ariel11 Should we update this to "or-later"? I noticed in the source repo that the file headers state "or later".

https://github.com/jnr/jnr-posix/blob/jnr-posix-3.1.4/src/main/java/jnr/posix/AixPOSIX.java

ariel11:
@capfei - interesting, it looks like only the base- and aix- files have that header, and I am not seeing it on the other java files.  But I would agree with you and update to "or-later."

capfei (approved):

capfei:
@ariel11 It is looking like it is going to be a weird day today. 😕 

@nellshamrell The validator is having issues again. It wasn't failing on version 3.1.4 until "only" was added to the identifier.

nellshamrell:
@capfei would you like me to go ahead and merge this?

capfei:
@nellshamrell Yes, please!

PR #14416, opened 2021-09-23
Update license

vuminhle:
**Type:** Incomplete

**Summary:**
Update license

**Details:**
Update the license according to the information here https://bitbucket.org/pypa/distlib/src/master/LICENSE.txt

**Resolution:**
Update license 

**Affected definitions**:
- [distlib 0.3.3](https://clearlydefined.io/definitions/pypi/pypi/-/distlib/0.3.3/0.3.3)


capfei:
This should be OTHER since the license is the full stack, not just the CRNI portion, and is not a full match to the Python-2.0 license.

@AE49 Do you agree?

PR #14477, opened 2021-09-24
utcstring 0.1.0

ariel11:
**Type:** Incorrect

**Summary:**
utcstring 0.1.0

**Details:**
Declared includes "MPL-2.0-no-copyleft-exception," which I think is wrong.  I think Exhibit B of the MPL license is just part of the template and not actually being applied. 

**Resolution:**
This looks like regular "MPL-2.0"

**Affected definitions**:
- [utcstring 0.1.0](https://clearlydefined.io/definitions/npm/npmjs/-/utcstring/0.1.0/0.1.0)

### DIFF
--- a/curations/npm/npmjs/-/utcstring.yaml
+++ b/curations/npm/npmjs/-/utcstring.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: utcstring
+  provider: npmjs
+  type: npm
+revisions:
+  0.1.0:
+    licensed:
+      declared: MPL-2.0


capfei:
@ariel11 I agree. The file headers do not include:

> This Source Code Form is “Incompatible With Secondary Licenses”, as defined by the Mozilla Public License, v. 2.0.

If this was present, then it would be the "no-copyleft-exception".

capfei (approved):

PR #15067, opened 2021-10-28
Update license

csim:
**Type:** Ambiguous

**Summary:**
Update license

**Details:**
License type not set

**Resolution:**
Added the Python-2.0 license

**Affected definitions**:
- [distlib 0.3.3](https://clearlydefined.io/definitions/pypi/pypi/-/distlib/0.3.3)

### DIFF
--- a/curations/pypi/pypi/-/distlib.yaml
+++ b/curations/pypi/pypi/-/distlib.yaml
@@ -14,4 +14,4 @@ revisions:
       declared: OTHER
   0.3.3:
     licensed:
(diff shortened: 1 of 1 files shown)


capfei:
Per the SPDX matching guidelines, the license is not a match to what they list as "Python-2.0" due to various differences in the license, particularly in the CNRI portion. Licenses that do not match a SPDX license identifier are curated OTHER. Closing PR since this component is already declared as OTHER.

PR #15199, opened 2021-11-05
xmldom 0.1.16

LandOfBliss:
**Type:** Missing

**Summary:**
xmldom 0.1.16

**Details:**
ClearlyDefined indicates license is LGPL
NPM license field indicate MIT
Open Source Project license is MIT OR LGPL-3.0-only: http://www.gnu.org/licenses/lgpl.html

**Resolution:**
Declared license is MIT OR LGPL-3.0-only

**Affected definitions**:
- [xmldom 0.1.16](https://clearlydefined.io/definitions/npm/npmjs/-/xmldom/0.1.16/0.1.16)

### DIFF
--- a/curations/npm/npmjs/-/xmldom.yaml
+++ b/curations/npm/npmjs/-/xmldom.yaml
@@ -5,7 +5,7 @@ coordinates:
 revisions:
   0.1.16:
     licensed:
-      declared: MIT OR LGPL-3.0-only
+      declared: LGPL-2.0-or-later
   0.1.19:
     licensed:
       declared: MIT OR LGPL-2.0-or-later
--- a/curations/npm/npmjs/-/xmldom.yaml
+++ b/curations/npm/npmjs/-/xmldom.yaml
@@ -3,6 +3,9 @@ coordinates:
   provider: npmjs
   type: npm
 revisions:
+  0.1.16:
+    licensed:
+      declared: MIT OR LGPL-3.0-only
   0.1.19:
     licensed:
       declared: MIT OR LGPL-2.0-or-later


AE49:
@LandOfBliss The version of this package was released in 2013 under "LGPL-3.0" and the MIT license was added to the repo more recently (Jan. 18, 2021). I would Declare this package / version as "LGPL-3.0-only" (no reference "or later" found for the LGPL license in the repo). @ariel11 Thoughts?

capfei:
@AE49 @LandOfBliss I would go with LGPL-2.0-only. I see in a later commit that the package.json is updated to LGPL 2.0 or MIT. This is before it is changed to MIT only:
https://github.com/xmldom/xmldom/commit/b7fd801dbff3f3d05a0d3f4fe3c2f0ef0f3f344f#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

What do you think?

ariel11:
It looks like the dual-license was added [Sep 10, 2013](https://github.com/jindw/xmldom/blob/afbf555e1b91179d3292fd6f46233799b541f044/LICENSE).  This version 0.1.16 is from May 4, 2013.  At that time, the license on the package.json is stated as: "licenses": [{"type": "LGPL","url": "http://www.gnu.org/licenses/lgpl.html"}].  Per our [curation guidelines (number 10 in troubleshooting)](https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md#troubleshooting), if the only license info is "LGPL" without a COPYING/LICENSE file, we will curate it as "LGPL-2.0-or-later."  So, I think this version would be curated "LGPL-2.0-or-later."  

Thoughts?  @capfei , @LandOfBliss , @AE49

AE49:
@ariel11 @capfei @LandOfBliss For this package version, the link in the package.json (https://github.com/xmldom/xmldom/blob/v0.1.16/package.json) has the full LGPL-3.0 license text - http://www.gnu.org/licenses/lgpl.html. This is why I was thinking we should curate as LGPL-3.0-only. Do you still think it should be LGPL-2.0-or-later?

ariel11:
@AE49 @LandOfBliss  @capfei  - we treat URLs like that as not indicating the _version_ of the license the project means to use, because that url just points to the most current/recent version of the L/GPL license.  In Section 10 of the curation guidelines, we have: "Note - http://www.gnu.org/licenses/gpl.txt is not a license version specific link."  And the url on the package.json is the same idea (http://www.gnu.org/licenses/lgpl.html).

ariel11 (approved):

PR #15260, opened 2021-11-10
org.slf4j/slf4j-nop/1.5.3

capfei:
**Type:** Missing

**Summary:**
org.slf4j/slf4j-nop/1.5.3

**Details:**
Add ICU

**Resolution:**
3 of the 4 source files have ICU license.

**Affected definitions**:
- [slf4j-nop 1.5.3](https://clearlydefined.io/definitions/maven/mavencentral/org.slf4j/slf4j-nop/1.5.3/1.5.3)

### DIFF
--- a/curations/maven/mavencentral/org.slf4j/slf4j-nop.yaml
+++ b/curations/maven/mavencentral/org.slf4j/slf4j-nop.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   1.5.3:
     licensed:
-      declared: ICU
+      declared: MIT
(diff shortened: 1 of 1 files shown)


ariel11:
@capfei - I don't know what I think on this one.  I see the 4 files and the ICU header, as you mentioned.  But then the `[pom ](https://repo1.maven.org/maven2/org/slf4j/slf4j-nop/1.5.3/slf4j-nop-1.5.3.pom)`points to http://www.slf4j.org, which says MIT - http://www.slf4j.org/license.html.  So maybe ICU is "discovered" and MIT is "declared"?

royaljust:
Agree with @ariel11 on this one.

ariel11 (approved):

PR #16581, opened 2021-12-22
Akka.Cluster.Tools 1.3.10

LandOfBliss:
**Type:** Missing

**Summary:**
Akka.Cluster.Tools 1.3.10

**Details:**
NuGet license field links to MIT
GitHub license is MIT: https://github.com/akkadotnet/akka.net/blob/v1.3.10/LICENSE

**Resolution:**
MIT

**Affected definitions**:
- [Akka.Cluster.Tools 1.3.10](https://clearlydefined.io/definitions/nuget/nuget/-/Akka.Cluster.Tools/1.3.10/1.3.10)

### DIFF
--- a/curations/nuget/nuget/-/Akka.Cluster.Tools.yaml
+++ b/curations/nuget/nuget/-/Akka.Cluster.Tools.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: Akka.Cluster.Tools
+  provider: nuget
+  type: nuget
+revisions:
+  1.3.10:
+    licensed:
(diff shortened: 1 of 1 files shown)


AE49:
@LandOfBliss The license link you point to is correct but it is the Apache-2.0 license text, not MIT and it looks like you meant Apache-2.0 just your notes say something different? Either way, Declare as Apache-2.0. @ariel11 Agree?

ariel11:
@AE49 - Agree with the Apache

AE49 (approved):

PR #16799, opened 2022-01-09
node-html-encoder 0.0.2

LandOfBliss:
**Type:** Missing

**Summary:**
node-html-encoder 0.0.2

**Details:**
ClearlyDefined package.json has no license info
NPM license field indicates NONE
GitHub has no license: https://github.com/minchenkov/node-html-encoder

**Resolution:**
NONE

**Affected definitions**:
- [node-html-encoder 0.0.2](https://clearlydefined.io/definitions/npm/npmjs/-/node-html-encoder/0.0.2/0.0.2)

### DIFF
--- a/curations/npm/npmjs/-/node-html-encoder.yaml
+++ b/curations/npm/npmjs/-/node-html-encoder.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: node-html-encoder
+  provider: npmjs
+  type: npm
+revisions:
+  0.0.2:
+    licensed:
+      declared: NONE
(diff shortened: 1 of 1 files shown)


capfei:
The only license info I can find is in `encoder.js`, which simply states "GPL". Since it looks like this is the main file for this package, I would say this should curated as GPL-1.0-or-later, since we have no other information.
https://github.com/minchenkov/node-html-encoder/blob/95ed4b3053d94bf37e4c96f45238e527f07850fe/lib/encoder.js#L5

@ae49 What do you think?

ariel11:
I agree because our curation guidelines say if the only license info is "GPL" we'll curate as "GPL-1.0-or-later."  And the only file in this library says "GPL."

ariel11 (approved):

PR #17082, opened 2022-01-21
Update libzmq license

jamill:
**Type:** Incorrect

**Summary:**
Update libzmq license

**Details:**
The component is labeled with a declared license of GPL-3.0, which appears to be inaccurate. The project declares it is licensed as LGPL.

I am not sure of the details of how the "declaration" was made from data sources - but thought this was a good starting point.

**Resolution:**
Update the declared license to be LGPL-3.0, as per:

https://github.com/zeromq/libzmq/tree/master#license

https://github.com/zeromq/libzmq/blob/master/COPYING.LESSER

**Affected definitions**:
- [libzmq 4097855ddaaa65ed7b5e8cb86d143842a594eebd](https://clearlydefined.io/definitions/git/github/zeromq/libzmq/4097855ddaaa65ed7b5e8cb86d143842a594eebd)

### DIFF
--- a/curations/git/github/zeromq/libzmq.yaml
+++ b/curations/git/github/zeromq/libzmq.yaml
@@ -7,3 +7,6 @@ revisions:
   2cb1240db64ce1ea299e00474c646a2453a8435b:
     licensed:
       declared: OTHER
+  4097855ddaaa65ed7b5e8cb86d143842a594eebd:
+    licensed:
+      declared: LGPL-3.0
--- a/curations/git/github/zeromq/libzmq.yaml
+++ b/curations/git/github/zeromq/libzmq.yaml
@@ -9,4 +9,4 @@ revisions:
       declared: OTHER
   4097855ddaaa65ed7b5e8cb86d143842a594eebd:
     licensed:
-      declared: LGPL-3.0
+      declared: OTHER


jamill:
Looking at the history of the curated data file, it appears the license label has switched from:

commit: ea0f027d75bbef3d25 - "LGPL-3.0"
commit: b2e3b9a5451d92b9af1e5e4f1ae8bfcdc514149d - "LGPL-3.0-or-later WITH OTHER"
commit: 9934985c5ae43cbf62b586247cd28b24f2ca10ee -"OTHER"

[libzmq.yaml history](https://github.com/clearlydefined/curated-data/commits/master/curations/git/github/zeromq/libzmq.yaml)

I am still not sure how the "declared" license tag is generated. Please advise if there is a more appropriate fix.

jamill:
Maybe the correct declaration is actually `LGPL-3.0-or-later`, as that string is actually in the list of discovered licenses.

capfei:
@jamill This should be OTHER since there is an exception at the bottom of the license. Since this exception does not have a SPDX exception ID, we would curate this as OTHER since it modifies that LGPL license (removes sections 4 and 5) and we no longer curate `WITH OTHER`.
```
           SPECIAL EXCEPTION GRANTED BY COPYRIGHT HOLDERS

As a special exception, copyright holders give you permission to link this
library with independent modules to produce an executable, regardless of
the license terms of these independent modules, and to copy and distribute
the resulting executable under terms of your choice, provided that you also
meet, for each linked independent module, the terms and conditions of
the license of that module. An independent module is a module which is not
derived from or based on this library. If you modify this library, you must
extend this exception to your version of the library.

Note: this exception relieves you of any obligations under sections 4 and 5
of this license, and section 6 of the GNU General Public License.
```

If you agree, I will update the PR.

jamill:
Thank you @capfei!

The current issue is that the "Declared" license is showing up as "GPL-3.0", which is incorrect, and I would like this to be more accurate.

If we update this to `OTHER`, what will the "Declared" license be? Will the "Declared" license be "Other"? If so, then I agree to your proposed update!

Thank you!

capfei:
@jamill Yes, the Declared will be "OTHER". I will update the PR. Thank you for confirming!

capfei (approved):

PR #17158, opened 2022-01-22
System.Text.Json2.0.0.11

LandOfBliss:
**Type:** Missing

**Summary:**
System.Text.Json2.0.0.11

**Details:**
NuGet license field is MIT
GitHub license is MIT: https://github.com/dotnet/runtime/blob/381372f6f12884caee5bb90f52649040827e19cc/LICENSE.TXT

**Resolution:**
MIT

**Affected definitions**:
- [System.Text.Json 2.0.0.11](https://clearlydefined.io/definitions/nuget/nuget/-/System.Text.Json/2.0.0.11/2.0.0.11)

### DIFF
--- a/curations/nuget/nuget/-/System.Text.Json.yaml
+++ b/curations/nuget/nuget/-/System.Text.Json.yaml
@@ -5,7 +5,7 @@ coordinates:
 revisions:
   2.0.0.11:
     licensed:
-      declared: MIT
+      declared: OTHER
   4.6.0:
     licensed:
       declared: MIT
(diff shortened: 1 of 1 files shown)


capfei:
I think this should be OTHER. The license link for this version points to a different location: https://github.com/ysharplanguage/FastJsonParser/blob/master/LICENSE.md.

The license looks like X11 with the copyright name updated: https://github.com/ysharplanguage/FastJsonParser/blob/fa30f3b76433148ff063caf5bbebb7f113f82a0f/LICENSE.md. On SPDX page, it doesn't show any of this text as editable.

@ariel11 What do you think?

ariel11:
Agree with OTHER because of the extra clause at the bottom of the license and the closest SPDX match (X11) doesn't have the differences in colored font, so it can't be different according to the SPDX matching guidelines. 

image

ariel11 (approved):

PR #17254, opened 2022-01-27
backports.shutil_which 3.5.2

gooooloo:
Type: Missing

Summary:
backports.shutil_which 3.5.2

Details:
Add MIT License

Resolution:
License Url:
https://github.com/minrk/backports.shutil_which/blob/3.5.2/LICENSE


capfei (approved):

capfei:

capfei:
Even though the setup.py states PSF as the license, the license file included with the package is MIT. Looking at the source repo, the `shutil_which.py` is PSF (imported module from the Python library) and everything else is MIT.

Agree with the declared being MIT and leaving PSF to fall under the "discovered" category.

PR #17295, opened 2022-01-28
coveralls.io 1.4.2

LandOfBliss:
**Type:** Missing

**Summary:**
coveralls.io 1.4.2

**Details:**
NuGet license field is: GPL-3.0-or-later
GitHub Readme license is GPL-3.0-or-later
https://github.com/coveralls-net/coveralls.net/blob/aac5494c82755ebdf63a53d460dec5aa19ea8eac/README.md

**Resolution:**
GPL-3.0-or-later

**Affected definitions**:
- [coveralls.io 1.4.2](https://clearlydefined.io/definitions/nuget/nuget/-/coveralls.io/1.4.2/1.4.2)

### DIFF
--- a/curations/nuget/nuget/-/coveralls.io.yaml
+++ b/curations/nuget/nuget/-/coveralls.io.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   1.4.2:
     licensed:
-      declared: GPL-3.0-or-later
+      declared: GPL-3.0-only
--- a/curations/nuget/nuget/-/coveralls.io.yaml
+++ b/curations/nuget/nuget/-/coveralls.io.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: coveralls.io
+  provider: nuget
+  type: nuget
+revisions:
+  1.4.2:
+    licensed:
+      declared: GPL-3.0-or-later


capfei:
@LandOfBliss I'm not seeing where it says they have the option for a later version. Am I missing it?
https://github.com/coveralls-net/coveralls.net/tree/fcf3fa860039da42dabe24043e393ec60588191c#license

> License
>
> Licensed under the GNU General Public License, Version 3.0 (the "License"); you may not use this work except in compliance with the License. You may obtain a copy of the License at:
>
> http://www.gnu.org/licenses/gpl-3.0.txt
> Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.

LandOfBliss:
@capfei I was referring to the Curation Guidelines #10: If the only license information is “GPL,” with no COPYING license file, code the license as “GPL-1.0-or-later.”

ariel11:
@LandOfBliss - The curation guideline you are referring to is when the _only_ license info provided by the author is literally just "GPL" and there is not a license or COPYING file.  Then we have the default curation of "GPL-1.0-or-later."  Otherwise, if a version of the license is specified (e.g., GPL-2.0, GPL-3.0), we determine if its "only" or "or-later" usually by looking at the license header on the files themselves. For this package, there is a license file that is GPL-3.0.  So, we now determine if its "GPL-3.0-only" or "GPL-3.0-or-later."  Since there doesn't appear to be a header indicating "or any later version," we would curate this as "GPL-3.0-only."

Cc: @capfei

capfei (approved):

PR #17441, opened 2022-02-03
tst 1.0.0

LandOfBliss:
**Type:** Missing

**Summary:**
tst 1.0.0

**Details:**
NPM license is MIT
GitHub package.json is MIT
GitHub license is ?: https://github.com/dy/tst/blob/v1.0.0/LICENSE
This appears to be a MIT like but not exact.

**Resolution:**
MIT

**Affected definitions**:
- [tst 1.0.0](https://clearlydefined.io/definitions/npm/npmjs/-/tst/1.0.0/1.0.0)

### DIFF
--- a/curations/npm/npmjs/-/tst.yaml
+++ b/curations/npm/npmjs/-/tst.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: tst
+  provider: npmjs
+  type: npm
+revisions:
+  1.0.0:
+    licensed:
+      declared: MIT
--- a/curations/npm/npmjs/-/tst.yaml
+++ b/curations/npm/npmjs/-/tst.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   1.0.0:
     licensed:
-      declared: MIT
+      declared: BSD-3-Clause


AE49:
@LandOfBliss I think we would Declare this as OTHER. I didn't get a good match on the DIFF tool to any SPDX license. @capfei Thoughts?

https://github.com/dy/tst/blob/v1.0.0/LICENSE
https://github.com/dy/tst/tree/v1.0.0#license

capfei:
@AE49 That tag is not for this package, but for `tape-modern`.
https://github.com/dy/tst/blob/v1.0.0/package.json#L2

Version 1.0.0 points to https://github.com/grahamlyons/tst as the source. It looks like this package changed owners. I can't find the history for version 1.0.0 of `tst` in the dy/tst repo. It starts with 2.0.0 in 2019.
https://github.com/dy/tst/commit/021cf4cc8bffe6796fb9b44165c67b935015ab45

I'm inclined to say to BSD-3-Clause per the curation guidelines. @ariel11 What do you think?

ariel11:
@capfei - yes, agree with BSD-3-Clause

PR #17865, opened 2022-02-17
License should not include NOASSERTION 

sophokles73:
**Type:** Incorrect

**Summary:**
License should not include NOASSERTION 

**Details:**
The license file in the META-INF folder gives no indication of NOASSERTION. Instead, it seems to be very clear about the fact that the component is BSD-2 licensed

**Resolution:**
Remove the NOASSERTION key word from the discovered license

**Affected definitions**:
- [postgresql 42.3.2](https://clearlydefined.io/definitions/maven/mavencentral/org.postgresql/postgresql/42.3.2/42.3.2)

### DIFF
--- a/curations/maven/mavencentral/org.postgresql/postgresql.yaml
+++ b/curations/maven/mavencentral/org.postgresql/postgresql.yaml
@@ -8,7 +8,7 @@ revisions:
     files:
       - attributions:
           - 'Copyright (c) 1997, PostgreSQL Global Development Group'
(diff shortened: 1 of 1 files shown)


capfei:
@sophokles73 Should the license file just be BSD-2-Clause instead of Apache-2.0 AND BSD-2-Clause?

capfei (approved):

capfei:
Thank you! Updated to BSD-2-Clause only.

PR #17875, opened 2022-02-17
pyzmq 19.0.2

ariel11:
**Type:** Incorrect

**Summary:**
pyzmq 19.0.2

**Details:**
In package there's a COPYING.BSD file, which is BSD-3-Clause.  That is the top "declared" license.  

**Resolution:**
There's also a COPYING.LESSER LGPL-3.0 file that is a "discovered" license, I think, as applies to just part of the package (cpython bindings).

**Affected definitions**:
- [pyzmq 19.0.2](https://clearlydefined.io/definitions/pypi/pypi/-/pyzmq/19.0.2/19.0.2)

### DIFF
--- a/curations/pypi/pypi/-/pyzmq.yaml
+++ b/curations/pypi/pypi/-/pyzmq.yaml
@@ -3,6 +3,9 @@ coordinates:
   provider: pypi
   type: pypi
 revisions:
+  19.0.2:
+    licensed:
+      declared: BSD-3-Clause
   22.0.2:
     licensed:
       declared: BSD-3-Clause AND LGPL-3.0-or-later


capfei:
@ariel11 I think this should be both BSD and LGPL. The CONTRIBUTING file states that the core code is LGPL and the non-core code is BSD.
https://github.com/zeromq/pyzmq/blob/v19.0.2/CONTRIBUTING.md#licensing-and-contributing-to-pyzmq

> Licensing and contributing to PyZMQ
>
> PyZMQ uses different licenses for different parts of the code.
>
> The 'core' of PyZMQ (located in zmq/core) is licensed under LGPLv3. This just means that if you make any changes to how that code works, you must release those changes under the LGPL. If you just use pyzmq, then you can use any license you want for your own code.
>
> We don't feel that the restrictions imposed by the LGPL make sense for the 'non-core' functionality in pyzmq (derivative code must also be LGPL or GPL), especially for examples and utility code, so we have relicensed all 'non-core' code under the more permissive BSD (specifically Modified BSD aka New BSD aka 3-clause BSD), where possible. This means that you can copy this code and build your own apps without needing to license your own code with the LGPL or GPL.

What do you think?

PR #18071, opened 2022-03-02
com.google.code.findbugs:jsr305 3.0.2

abyss638:
**Type:** Incorrect

**Summary:**
License is incorrect in Maven Central

**Details:**
The correct license is here https://github.com/findbugsproject/findbugs/blob/master/findbugs/licenses/LICENSE-jsr305.txt

**Resolution:**
License is updated

**Affected definitions**:
- [jsr305 3.0.2](https://clearlydefined.io/definitions/sourcearchive/mavencentral/com.google.code.findbugs/jsr305/3.0.2/3.0.2)

### DIFF
--- a/curations/sourcearchive/mavencentral/com.google.code.findbugs/jsr305.yaml
+++ b/curations/sourcearchive/mavencentral/com.google.code.findbugs/jsr305.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   3.0.2:
     licensed:
-      declared: BSD-2-Clause
+      declared: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


capfei:
@abyss638 If you agree, I would like to update this to BSD-3-Clause. Looking at the [source repo](https://code.google.com/archive/p/jsr-305/) listed in the pom, it points to BSD-3-Clause: http://opensource.org/licenses/BSD-3-Clause.

Also, the license information your provided above states the New BSD license, which is BSD-3-Clause. The link, https://opensource.org/licenses/bsd-license.php, used originally pointed to 3-Clause, but now points to 2-Clause so we do not use it determine version.

Let me know if you agree and I will update the PR.

abyss638:
@capfei I agree with BSD-3-Clause

capfei (approved):

capfei:
Forgot to include this link where the package authors state that the license listed in the pom (Apache) is not correct and that this is BSD.
https://github.com/findbugsproject/findbugs/issues/128

PR #18274, opened 2022-03-08
junit-addons/junit-addons 1.4

LandOfBliss:
**Type:** Missing

**Summary:**
junit-addons/junit-addons 1.4

**Details:**
Maven pom has no license info
Jar includes license which is OTHER (The JUnit-addons Software License, Version 1.0
 (based on the Apache Software License, Version 1.1)

**Resolution:**
OTHER

**Affected definitions**:
- [junit-addons 1.4](https://clearlydefined.io/definitions/maven/mavencentral/junit-addons/junit-addons/1.4/1.4)

### DIFF
--- a/curations/maven/mavencentral/junit-addons/junit-addons.yaml
+++ b/curations/maven/mavencentral/junit-addons/junit-addons.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: junit-addons
+  namespace: junit-addons
+  provider: mavencentral
+  type: maven
+revisions:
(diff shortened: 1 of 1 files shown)


capfei (approved):

capfei:
Updated to `Apache-1.1`. Diff tool shows 72.2% match, but when you look at what was changed, it is the text that is marked as editable per SPDX matching guidelines.
https://spdx.org/licenses/Apache-1.1.html

PR #18459, opened 2022-03-17
boost_thread-vc141 1.68.0

LandOfBliss:
**Type:** Missing

**Summary:**
boost_thread-vc141 1.68.0

**Details:**
Nuget license is BSL-1.0
GitHub license at time of package was Apache-2.0 but later changed to BSL-1.0

**Resolution:**
BSL-1.0

**Affected definitions**:
- [boost_thread-vc141 1.68.0](https://clearlydefined.io/definitions/nuget/nuget/-/boost_thread-vc141/1.68.0/1.68.0)

### DIFF
--- a/curations/nuget/nuget/-/boost_thread-vc141.yaml
+++ b/curations/nuget/nuget/-/boost_thread-vc141.yaml
@@ -3,6 +3,9 @@ coordinates:
   provider: nuget
   type: nuget
 revisions:
+  1.68.0:
+    licensed:
+      declared: BSL-1.0
   1.69.0:
     licensed:
       declared: BSL-1.0


AE49:
@LandOfBliss NuGet license is pointing to the Master which is the most recent license updated August 2021.

This version 1.68.0 is under Apache-2.0 and we would Declare license at time package was released - 

https://github.com/sergey-shandar/getboost/blob/fa5fd42aad47057ee364ceaf086ea541d9b53ef2/LICENSE. Do you agree we should Declare Apache-2.0?

AE49:
@capfei Do you agree with Apache-2.0?

ariel11:
@AE49 - Per https://github.com/sergey-shandar/getboost/issues/64, we are using BSL-1.0 for the declared license.

PR #18551, opened 2022-03-22
org.jruby/jruby 1.7.0

ariel11:
**Type:** Incorrect

**Summary:**
org.jruby/jruby 1.7.0

**Details:**
Not sure where the tooling is pulling LGPL-3.0.  The LICENSE and NOTICE both reference Apache-2.0 as the license.

**Resolution:**
Declared = Apache-2.0

**Affected definitions**:
- [jruby 1.7.0](https://clearlydefined.io/definitions/maven/mavencentral/org.jruby/jruby/1.7.0/1.7.0)

### DIFF
--- a/curations/maven/mavencentral/org.jruby/jruby.yaml
+++ b/curations/maven/mavencentral/org.jruby/jruby.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: jruby
+  namespace: org.jruby
+  provider: mavencentral
+  type: maven
+revisions:
+  1.7.0:
+    licensed:
+      declared: Apache-2.0


ariel11:
@capfei - we also located the source repo, and that says Jruby is tri-licensed - https://github.com/jruby/jruby/blob/1.7.0/COPYING.  It may be that the Apache LICENSE/NOTICE files in the package are actually only applicable to a portion of the package.  What are your thoughts on this one?

capfei (approved):

capfei:
@ariel11 I revisited this one after thinking about it some more and realized that I wasn't able to locate this license.txt. Reading through issues, I think that the Apache license was being included in error (seen this kind of issue before).

I think this should be corrected to the tri-license listed in the source repo and in the sources.jar.
`CPL-1.0 OR GPL-2.0-or-later OR LGPL-2.1-or-later` If you agree, I will correct this one.

PR #18603, opened 2022-03-23
Helpshift 7.6.3

LandOfBliss:
**Type:** Missing

**Summary:**
Helpshift 7.6.3

**Details:**
GitHub license is Apache-2.0: https://github.com/helpshift/helpshift-api

**Resolution:**
Apache-2.0

**Affected definitions**:
- [Helpshift 7.6.3](https://clearlydefined.io/definitions/pod/cocoapods/-/Helpshift/7.6.3/7.6.3)

### DIFF
--- a/curations/pod/cocoapods/-/Helpshift.yaml
+++ b/curations/pod/cocoapods/-/Helpshift.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   7.6.3:
     licensed:
-      declared: Apache-2.0
+      declared: OTHER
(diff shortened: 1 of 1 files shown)


capfei:
I think this should be OTHER since the podspec states the license is "Commerical" and points to http://www.helpshift.com/terms/.

https://github.com/CocoaPods/Specs/blob/master/Specs/2/a/d/Helpshift/7.6.3/Helpshift.podspec.json#L5-L7

@AE49 What do you think?

AE49:
@capfei @LandOfBliss - I agree with OTHER for the commercial license.

AE49:
@capfei @LandOfBliss I agree with OTHER.

PR #18615, opened 2022-03-24
Update jruby.yaml

capfei:


capfei:
@fossygirl, @jeffwilcox Can one of you merge this? It is to fix https://github.com/clearlydefined/curated-data/pull/18551#issuecomment-1077815549, which I merged incorrectly as Apache.

The validator bot does not like multiple ORs so this needs to be force merged.

PR #18834, opened 2022-04-04
xpath.js1.0.7

AE49:
**Type:** Incorrect

**Summary:**
xpath.js1.0.7

**Details:**
Fixing curation to MIT

**Resolution:**
https://github.com/yaronn/xpath.js/pull/22

**Affected definitions**:
- [xpath.js 1.0.7](https://clearlydefined.io/definitions/npm/npmjs/-/xpath.js/1.0.7/1.0.7)

### DIFF
--- a/curations/npm/npmjs/-/xpath.js.yaml
+++ b/curations/npm/npmjs/-/xpath.js.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   1.0.7:
     licensed:
-      declared: CC-BY-SA-2.0
+      declared: MIT


capfei:
You can make a case that this should be left as CC-BY-SA-2.0 since the README states that it is based on code that is the same license and the author stated it is CC-BY-SA-2.0 back in Dec 2015. This package was published in 2016 and the license changed to MIT in 2017.
https://github.com/yaronn/xpath.js#xpathjs
https://github.com/yaronn/xpath.js/issues/15

@ariel11 What do you think?

ariel11:
Agree at the point in time the package was published it was still CC.

PR #18960, opened 2022-04-07
py-ms 2.4.0

LandOfBliss:
**Type:** Missing

**Summary:**
py-ms 2.4.0

**Details:**
PyPi indicates GPLv3
ClearlyDefined PACKAGE-INFO indicates GPLv3
Curation guidelines #10

**Resolution:**
GPL-3.0-or-later

**Affected definitions**:
- [py-ms 2.4.0](https://clearlydefined.io/definitions/pypi/pypi/-/py-ms/2.4.0/2.4.0)

### DIFF
--- a/curations/pypi/pypi/-/py-ms.yaml
+++ b/curations/pypi/pypi/-/py-ms.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   2.4.0:
     licensed:
-      declared: GPL-3.0-or-later
+      declared: GPL-3.0-only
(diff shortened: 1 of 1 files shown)


capfei:
@LandOfBliss Can you point me to where it states "or later"? I checked several files but could not find anything that states it.

LandOfBliss:
@capfei I did not find evidence of "only" or "or later".  So, followed the curation guidelines #10.

capfei:
@LandOfBliss Ok, then without any references to "or later" then this should "GPL-3.0-only" since it is stated to be GPLv3. The only time we would default "or later" is if the only license information we have is just "GPL" and no version, which we would then use GPL-1.0-or-later.

Do you agree?

ariel11:
Agree with "GPL-3.0-only."

capfei (approved):

PR #18889, opened 2022-04-05
ckeditor4-react 2.1.1

clearlydefinedbot:
**Type:** Missing

**Summary:**
ckeditor4-react 2.1.1

**Details:**
Add GPL-3.0-only License

**Resolution:**
License Url: 
https://github.com/ckeditor/ckeditor4-react/blob/master/LICENSE.md

Description:

Pull request generated by Microsoft tooling.

**Affected definitions**:
- [ckeditor4-react 2.1.1](https://clearlydefined.io/definitions/npm/npmjs/-/ckeditor4-react/2.1.1)

PR #19585, opened 2022-05-09
Update ckeditor4-react.yaml

capfei:
Trying to work around validator bot issue


capfei:
tried to fix #18889, #19584

PR #19602, opened 2022-05-11
Declared License Missing

Prashant30011992:
**Type:** Missing

**Summary:**
Declared License Missing

**Details:**
Declared License should be LGPL 3.0 and GPL 3.0 instead of NOASSERTION 
License Path:
https://github.com/luisgoncalves/xades4j/blob/v1.3.2/license.txt

https://github.com/luisgoncalves/xades4j/blob/v1.3.2/gpl-3.0.txt

**Resolution:**
License Path:
https://github.com/luisgoncalves/xades4j/blob/v1.3.2/license.txt

https://github.com/luisgoncalves/xades4j/blob/v1.3.2/gpl-3.0.txt

**Affected definitions**:
- [xades4j 388c7ea01cbe7a2514436bf5bc4ba2b9b28d8348](https://clearlydefined.io/definitions/git/github/luisgoncalves/xades4j/388c7ea01cbe7a2514436bf5bc4ba2b9b28d8348/388c7ea01cbe7a2514436bf5bc4ba2b9b28d8348)

### DIFF
--- a/curations/git/github/luisgoncalves/xades4j.yaml
+++ b/curations/git/github/luisgoncalves/xades4j.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   388c7ea01cbe7a2514436bf5bc4ba2b9b28d8348:
     licensed:
-      declared: LGPL-3.0-only AND GPL-3.0-only
+      declared: LGPL-3.0-or-later
--- a/curations/git/github/luisgoncalves/xades4j.yaml
+++ b/curations/git/github/luisgoncalves/xades4j.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: xades4j
+  namespace: luisgoncalves
+  provider: github
+  type: git
+revisions:
+  388c7ea01cbe7a2514436bf5bc4ba2b9b28d8348:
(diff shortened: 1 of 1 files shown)


capfei:
I think this should be LGPL-3.0-or-later for a couple of reasons. First, the file headers state LGPL 3.0, not both LGPL and GPL. Also, LGPL 3.0 can include the GPL 3.0 license or not. From my understanding, LGPL 3.0 is an addition to GPL 3.0 so it can/not with GPL 3.0 license text but is still LGPL 3.0.
https://github.com/spdx/license-list-XML/issues/956

SPDX has the GPL 3.0 license text as optional for LGPL 3.0.
https://spdx.org/licenses/LGPL-3.0-or-later.html

@ariel11 What do you think?

ariel11:
@capfei - yes, I agree with LGPL-3.0-or-later (example file: https://github.com/luisgoncalves/xades4j/blob/388c7ea01cbe7a2514436bf5bc4ba2b9b28d8348/src/main/java/xades4j/XAdES4jException.java).

capfei (approved):

PR #19662, opened 2022-05-13
org.tmatesoft.sqljet/sqljet/1.1.10

capfei:
**Type:** Missing

**Summary:**
org.tmatesoft.sqljet/sqljet/1.1.10

**Details:**
Pom states "GPL (with dual licensing option)", but I could not find any dual license information. The license file states GPL 2.0  only and the source file headers are the same.

**Resolution:**
GPL-2.0-only

**Affected definitions**:
- [sqljet 1.1.10](https://clearlydefined.io/definitions/maven/mavencentral/org.tmatesoft.sqljet/sqljet/1.1.10/1.1.10)

### DIFF
--- a/curations/maven/mavencentral/org.tmatesoft.sqljet/sqljet.yaml
+++ b/curations/maven/mavencentral/org.tmatesoft.sqljet/sqljet.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: sqljet
+  namespace: org.tmatesoft.sqljet
+  provider: mavencentral
+  type: maven
+revisions:
+  1.1.10:
+    licensed:
+      declared: GPL-2.0-only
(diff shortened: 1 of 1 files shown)


AE49:
@capfei I think it could be GPL-2.0-or-later OR OTHER (for the commercial license option) - https://www.openhub.net/p/sqljet.

The project homepage is asking for sign-in and I think that is where the commercial license details reside.

See also bottom of this page on the dual-licensing - 

https://java-library-catalog.blogspot.com/

ariel11:
@capfei  and @AE49 - I agree with adding "OR OTHER" as the license header has the highlighted part.  However, I would do GPL-2.0-only.  So the whole thing I think would be "GPL-2.0-only OR OTHER."  Thoughts?

image

AE49 (approved):

PR #20334, opened 2022-06-28
Source Location was missing and the declared license was incorrect.

SwadhinMangaraj:
**Type:** Missing

**Summary:**
Source Location was missing and the declared license was incorrect.

**Details:**
Source location URL was missing and declared license is incorrect.

**Resolution:**
Filled in the source location URL as per https://www.nuget.org/packages/IronPython.StdLib/2.7.11 --> https://ironpython.net/ --> https://github.com/IronLanguages/ironpython2/tree/c02d86cea702fd6d253e8fb988187789f7a631b7 and the license as per https://github.com/IronLanguages/ironpython2/blob/c02d86.

**Affected definitions**:
- [IronPython.StdLib 2.7.11](https://clearlydefined.io/definitions/nuget/nuget/-/IronPython.StdLib/2.7.11/2.7.11)

### DIFF
--- a/curations/nuget/nuget/-/IronPython.StdLib.yaml
+++ b/curations/nuget/nuget/-/IronPython.StdLib.yaml
@@ -7,8 +7,16 @@ revisions:
     licensed:
       declared: OTHER
   2.7.11:
+    described:
+      sourceLocation:
+        name: ironpython2
+        namespace: ironlanguages
+        provider: github
+        revision: c02d86cea702fd6d253e8fb988187789f7a631b7
+        type: git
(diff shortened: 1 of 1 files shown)


capfei:
I agree with the source location information but the declared license for the package points to http://docs.python.org/license.html, which we curate as OTHER due to differences in what SPDX has as the Python-2.0 license.

The latest version of this package also includes the Python license, so I don't think it is an error.
https://www.nuget.org/packages/IronPython.StdLib/3.4.0-beta1/License

@AE49 What do you think?

AE49:
@capfei Agree with OTHER

capfei (approved):

PR #20557, opened 2022-07-13
EPPlus4.0.1.1

LandOfBliss:
**Type:** Missing

**Summary:**
EPPlus4.0.1.1

**Details:**
Nuspec license link is:
http://epplus.codeplex.com/license which is no longer accessible.  No license info found.

**Resolution:**
NONE

**Affected definitions**:
- [EPPlus 4.0.1.1](https://clearlydefined.io/definitions/nuget/nuget/-/EPPlus/4.0.1.1/4.0.1.1)

### DIFF
--- a/curations/nuget/nuget/-/EPPlus.yaml
+++ b/curations/nuget/nuget/-/EPPlus.yaml
@@ -3,6 +3,9 @@ coordinates:
   provider: nuget
   type: nuget
 revisions:
+  4.0.1.1:
+    licensed:
+      declared: NONE
   4.0.4:
     licensed:
       declared: LGPL-2.1-or-later
--- a/curations/nuget/nuget/-/EPPlus.yaml
+++ b/curations/nuget/nuget/-/EPPlus.yaml
@@ -5,13 +5,13 @@ coordinates:
 revisions:
   4.0.1.1:
     licensed:
-      declared: NONE
+      declared: LGPL-2.1-or-later
   4.0.4:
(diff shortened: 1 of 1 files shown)


capfei:
Downloaded archive and file headers state LGPL-2.1-or-later.
```
/*******************************************************************************
 * You may amend and distribute as you like, but don't remove this header!
 *
 * EPPlus provides server-side generation of Excel 2007/2010 spreadsheets.
 * See http://www.codeplex.com/EPPlus for details.
 *
 * Copyright (C) 2011  Jan Källman
 *
 * This library is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License as published by the Free Software Foundation; either
 * version 2.1 of the License, or (at your option) any later version.

 * This library is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  
 * See the GNU Lesser General Public License for more details.
 *
 * The GNU Lesser General Public License can be viewed at http://www.opensource.org/licenses/lgpl-license.php
 * If you unfamiliar with this license or have questions about it, here is an http://www.gnu.org/licenses/gpl-faq.html
 *
 * All code and executables are provided "as is" with no warranty either express or implied. 
 * The author accepts no liability for any damage or loss of business that this product may cause.
```

PR #20692, opened 2022-07-22
tomcat/jasper-runtime 5.5.12

LandOfBliss:
**Type:** Missing

**Summary:**
tomcat/jasper-runtime 5.5.12

**Details:**
License is Apache-2.0: https://github.com/sonatype-nexus-community/search-maven-org/blob/master/LICENSE.txt

**Resolution:**
Apache-2.0

**Affected definitions**:
- [jasper-runtime 5.5.12](https://clearlydefined.io/definitions/maven/mavencentral/tomcat/jasper-runtime/5.5.12/5.5.12)

### DIFF
--- a/curations/maven/mavencentral/tomcat/jasper-runtime.yaml
+++ b/curations/maven/mavencentral/tomcat/jasper-runtime.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: jasper-runtime
+  namespace: tomcat
+  provider: mavencentral
+  type: maven
+revisions:
+  5.5.12:
+    licensed:
+      declared: Apache-2.0
--- a/curations/maven/mavencentral/tomcat/jasper-runtime.yaml
+++ b/curations/maven/mavencentral/tomcat/jasper-runtime.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   5.5.12:
     licensed:
-      declared: Apache-2.0
+      declared: NONE


capfei:
@LandOfBliss  I'm not finding any license info. Where is that link in the package?

LandOfBliss:
> @LandOfBliss I'm not finding any license info. Where is that link in the package?

@capfei I didn't find it in the package but here: https://mvnrepository.com/artifact/tomcat/jasper-runtime

capfei:
That isn't a reliable source for license information. @ariel11 What do you think? I'm not finding any license info. NONE?

ariel11:
@capfei  and @LandOfBliss - Yes, mvnrepository site is not considered a canonical source for license info.  We do have that curation guideline that says if a different package version has license info we can use that for curation purposes.  I checked the most recent package version, but it doesn't loook like the .pom has any license info and does not look like they are publishing a sources jar?

capfei (approved):

ariel11:
@capfei - For something else, I was looking at this version of this package: https://repo1.maven.org/maven2/tomcat/jasper-runtime/5.5.9/ and it does have a sources.jar and we can see the Apache-2.0 code comments in the file headers.  I think we can use the curation guideline to update this package version to Apache-2.0 based on this other package version.  Thoughts?

capfei:
@ariel11 Yes, we can update it. I think I need to write a script to search through and look for versions with a sources.jar. Clicking through 20 versions to find the only 2 with a sources.jar is too frustrating and time consuming. 😠

PR #21167, opened 2022-08-12
Declared License is missing

SwadhinMangaraj:
**Type:** Missing

**Summary:**
Declared License is missing

**Details:**
Declared license is missing possibly because ClearDefined tool could not interpret the license.

**Resolution:**
This component is distributed under Python Imaging Software License which is an MIT-style permissive license. Filled in the correct as per https://github.com/python-pillow/Pillow/blob/58acec3312fb8671c9d84829197e1c8150085589/LICENSE.

**Affected definitions**:
- [pillow 58acec3312fb8671c9d84829197e1c8150085589](https://clearlydefined.io/definitions/git/github/python-pillow/pillow/58acec3312fb8671c9d84829197e1c8150085589/58acec3312fb8671c9d84829197e1c8150085589)

### DIFF
--- a/curations/git/github/python-pillow/pillow.yaml
+++ b/curations/git/github/python-pillow/pillow.yaml
@@ -6,7 +6,7 @@ coordinates:
 revisions:
   4f6145655b6fa35ec8dd3600041418087e399758:
     licensed:
-      declared: HPND
+      declared: MIT-CMU
   58acec3312fb8671c9d84829197e1c8150085589:
     licensed:
-      declared: OTHER
+      declared: MIT-CMU
--- a/curations/git/github/python-pillow/pillow.yaml
+++ b/curations/git/github/python-pillow/pillow.yaml
@@ -7,3 +7,6 @@ revisions:
   4f6145655b6fa35ec8dd3600041418087e399758:
     licensed:
       declared: HPND
+  58acec3312fb8671c9d84829197e1c8150085589:
+    licensed:
+      declared: OTHER


capfei:
@ariel11 Even though it says HPND, it is not a match. There are some key differences that match to MIT-CMU.
Match to MIT-CMU:
image

Match to HPND:
image

HPND does not contain the first paragraph:
>By obtaining, using, and/or copying this software and/or its associated
documentation, you agree that you have read, understood, and will comply
with the following terms and conditions:

Also, "associated" is not used in HPND:
> Permission to use, copy, modify, and distribute this software and its
_**associated**_ documentation for any purpose

In the SPDX license list, Pillow is listed as an example of MIT-CMU: https://spdx.org/licenses/MIT-CMU.html

We have talked about this one in the past and stayed with HPND because that is what they state and MIT-CMU didn't exist when Pillow started using this license. However, I think we should start using MIT-CMU since that looks to be the now correct license identifier and to follow the matching guidelines.

What do you think?

capfei:
@ariel11 Actually, I just realized that Pillow has the "prior" for "written prior permission" but the SPDX does not, but they link to this license as an example. 

I will look into this some more and may have to request the word be added as optional.

capfei:
Filed an issue: https://github.com/spdx/license-list-XML/issues/1578. Will update when I get a response.

capfei:
Seems there is not an issue with adding "prior" as an optional tag, so this makes Pillow's license a match to MIT-CMU. Created a PR to add: https://github.com/spdx/license-list-XML/pull/1581

ariel11:
@capfei - interesting.  Yes, looking at it from a purely SPDX matching guidelines perspective, I think this would be "OTHER."  What are your thoughts on that?

capfei:
@ariel11 SPDX merged my edit to add "prior" as an optional tag today so, to me, this now makes it a match to MIT-CMU. Unfortunately, it won't show up until the next release so matching will still be at 91.5% instead of higher. Every thing else that isn't matching on MIT-CMU are editable or optional ("the copyright holders" and "the").
https://spdx.org/licenses/MIT-CMU.html

ariel11:
@capfei - OK, yes - if it will meet the matching guidelines at the next SPDX release/update, I think it's OK for us to use the MIT-CMU SPDX now...

capfei (approved):

PR #21549, opened 2022-09-22
pyside2 5.15.2.1 license correction

copernicat:
**Type:** Incorrect

**Summary:**
license correction

**Details:**
package is alternatively licensed

**Resolution:**
Attempts to correct declared license. Cannot multi-select more than two licenses, but the actual license is SPDX-License-Identifier: LicenseRef-Qt-Commercial OR LGPL-3.0-only OR GPL-2.0-only OR GPL-3.0-only

**Affected definitions**:
- [pyside2 5.15.2.1](https://clearlydefined.io/definitions/pypi/pypi/-/pyside2/5.15.2.1/5.15.2.1)

### DIFF
--- a/curations/pypi/pypi/-/pyside2.yaml
+++ b/curations/pypi/pypi/-/pyside2.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: pyside2
+  provider: pypi
+  type: pypi
+revisions:
+  5.15.2.1:
+    licensed:
+      declared: GPL-3.0-only OR LGPL-3.0-only
--- a/curations/pypi/pypi/-/pyside2.yaml
+++ b/curations/pypi/pypi/-/pyside2.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   5.15.2.1:
     licensed:
-      declared: GPL-3.0-only OR LGPL-3.0-only
+      declared: GPL-2.0 only OR GPL-3.0-only OR LGPL-3.0-only OR OTHER


capfei:
@ariel11 Looking at the license info in the files, as well as the source, the license states:
```
## $QT_BEGIN_LICENSE:LGPL$
## Commercial License Usage
## Licensees holding valid commercial Qt licenses may use this file in
## accordance with the commercial license agreement provided with the
## Software or, alternatively, in accordance with the terms contained in
## a written agreement between you and The Qt Company. For licensing terms
## and conditions see https://www.qt.io/terms-conditions. For further
## information use the contact form at https://www.qt.io/contact-us.
##
## GNU Lesser General Public License Usage
## Alternatively, this file may be used under the terms of the GNU Lesser
## General Public License version 3 as published by the Free Software
## Foundation and appearing in the file LICENSE.LGPL3 included in the
## packaging of this file. Please review the following information to
## ensure the GNU Lesser General Public License version 3 requirements
## will be met: https://www.gnu.org/licenses/lgpl-3.0.html.
##
## GNU General Public License Usage
## Alternatively, this file may be used under the terms of the GNU
## General Public License version 2.0 or (at your option) the GNU General
## Public license version 3 or any later version approved by the KDE Free
## Qt Foundation. The licenses are as published by the Free Software
## Foundation and appearing in the file LICENSE.GPL2 and LICENSE.GPL3
## included in the packaging of this file. Please review the following
## information to ensure the GNU General Public License requirements will
## be met: https://www.gnu.org/licenses/gpl-2.0.html and
## https://www.gnu.org/licenses/gpl-3.0.html.
```

If am reading correctly, this should be `GPL-2.0-only OR GPL-3.0-or-later OR LGPL-3.0-only OR OTHER`. This is right? 🤔

capfei:
Will update the same to #21550 since it has the same license info.

copernicat:
Hi @capfei , thank you for checking this one carefully! I should have explained better: Attempts to correct declared license. Cannot multi-select more than two licenses, but the actual license is SPDX-License-Identifier: LicenseRef-Qt-Commercial OR LGPL-3.0-only OR GPL-2.0-only OR GPL-3.0-only
I chose GPL-3.0 only because it has additional requirements compared to the GPL-2.0. 
It would be a nice future feature upgrade to be able to select more than two licenses.

ariel11:
@capfei and @copernicat - Interesting one.  I’d do OTHER (for commercial license option, per the curation guidelines) OR LGPL-3.0-only OR GPL-2.0-only OR GPL-3.0-only (because the “or later” is conditioned upon approval by the KDE Free QT Foundation, so it's not really a license option at this point in time.  So, "OTHER OR LGPL-3.0-only OR GPL-2.0 only OR GPL-3.0-only," is where I come out. 

Thoughts? :)

capfei (approved):

capfei:
Thank you @copernicat @ariel11 🎉

PR #21550, opened 2022-09-22
shiboken2 5.15.2.1 license correction

copernicat:
**Type:** Incorrect

**Summary:**
license correction

**Details:**
package is alternatively licensed

**Resolution:**
Attempts to correct declared license. Cannot multi-select more than two licenses, but the actual license is SPDX-License-Identifier: LicenseRef-Qt-Commercial OR LGPL-3.0-only OR GPL-2.0-only OR GPL-3.0-only

**Affected definitions**:
- [shiboken2 5.15.2.1](https://clearlydefined.io/definitions/pypi/pypi/-/shiboken2/5.15.2.1/5.15.2.1)

### DIFF
--- a/curations/pypi/pypi/-/shiboken2.yaml
+++ b/curations/pypi/pypi/-/shiboken2.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   5.15.2.1:
     licensed:
-      declared: GPL-3.0-only OR LGPL-3.0-only
+      declared: GPL-2.0-only OR GPL-3.0-only OR LGPL-3.0-only OR OTHER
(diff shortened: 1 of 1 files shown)


capfei:
Will update the same as #21549 since it has the same license info.

copernicat:
Hi @capfei , thank you for checking this one carefully! I should have explained better: Attempts to correct declared license. Cannot multi-select more than two licenses, but the actual license is SPDX-License-Identifier: LicenseRef-Qt-Commercial OR LGPL-3.0-only OR GPL-2.0-only OR GPL-3.0-only
I chose GPL-3.0 only because it has additional requirements compared to the GPL-2.0.
It would be a nice future feature upgrade to be able to select more than two licenses.

capfei (approved):

PR #21614, opened 2022-09-28
Adding declarations for all versions

marobbins:
Identical license is used for all versions listed on nuget


capfei:
Thank you for the contribution! 😄 

One ask: if possible, please try to break up large changes in to 10 item PRs. This makes it easier for us to review and breaks up the work so it can be divided, as needed.

PR #21802, opened 2022-10-10
Correct Declared License field

copernicat:
**Type:** Incorrect

**Summary:**
Correct Declared License field

**Details:**
The OS package's README file explains that some files are under LGPL-3.0-only and some are under BSD-3-Clause. See https://clearlydefined.io/file/8ec1f2ce30b9e6627f93500bd6036c11e1a7bc1b46621aa5058d8f2a5c0d2133. I tried to update it to LGPL-3.0-only AND BSD-3-Clause, but the UI wouldn't let me

**Resolution:**
Update Declared License field

**Affected definitions**:
- [pyzmq 23.2.1](https://clearlydefined.io/definitions/pypi/pypi/-/pyzmq/23.2.1/23.2.1)

### DIFF
--- a/curations/pypi/pypi/-/pyzmq.yaml
+++ b/curations/pypi/pypi/-/pyzmq.yaml
@@ -9,3 +9,6 @@ revisions:
   22.0.2:
     licensed:
       declared: BSD-3-Clause
+  23.2.1:
+    licensed:
+      declared: NOASSERTION AND BSD-3-Clause
--- a/curations/pypi/pypi/-/pyzmq.yaml
+++ b/curations/pypi/pypi/-/pyzmq.yaml
@@ -11,4 +11,4 @@ revisions:
       declared: BSD-3-Clause
   23.2.1:
     licensed:
-      declared: NOASSERTION AND BSD-3-Clause
+      declared: BSD-3-Clause


capfei:
@ariel11 Can I get your opinion here? Looking at past PRs, it seems we have gone back and forth between both and BSD-3-Clause only.
https://github.com/clearlydefined/curated-data/pulls?q=is%3Apr+sort%3Aupdated-desc+pyzmq+is%3Aclosed

Should we continue with that or update to reflect what the author is declaring, even though LGPL applies to a single folder?
https://github.com/zeromq/pyzmq/issues/1610
https://github.com/zeromq/pyzmq/tree/v23.2.1/zmq/backend/cython  <-- this one 

ℹ️ __Note:__ It would not be LGPL, but OTHER, since there is an exception at the bottom of the COPYING.LESSER file, which is does not match the classpath exception.
https://github.com/zeromq/pyzmq/blob/v23.2.1/COPYING.LESSER#L169-L181

ariel11:
Thanks, @capfei!  I think based on the RELICENSE file https://github.com/zeromq/pyzmq/tree/v23.2.1/RELICENSE I would say the "declared" license (as ClearlyDefined defines it) is BSD-3-Clause and the LGPL for cython is "discovered."  However, I am curious what others think, so let's wait a minute before merging.  Thanks!

ariel11:
@capfei - we discussed and agree the declared here would be BSD-3-Clause.  Thanks!

capfei (approved):

PR #21923, opened 2022-10-17
Icu4c.Win.Min59.1.7

AE49:
**Type:** Incomplete

**Summary:**
Icu4c.Win.Min59.1.7

**Details:**
Declare ICU

**Resolution:**
https://github.com/sillsdev/icu4c-svn/blob/release-59-1/license.html

**Affected definitions**:
- [Icu4c.Win.Min 59.1.7](https://clearlydefined.io/definitions/nuget/nuget/-/Icu4c.Win.Min/59.1.7/59.1.7)

### DIFF
--- a/curations/nuget/nuget/-/Icu4c.Win.Min.yaml
+++ b/curations/nuget/nuget/-/Icu4c.Win.Min.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: Icu4c.Win.Min
+  provider: nuget
+  type: nuget
+revisions:
+  59.1.7:
+    licensed:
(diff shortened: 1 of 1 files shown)


capfei:
@AE49 I think this should be OTHER since the license text does not match ICU (refers to data files) and it does not match the Unicode-DFS-2016 since it is missing the "NOTICE TO USER" paragraph.
https://github.com/sillsdev/icu4c-svn/blob/release-59-1/LICENSE

AE49:
@capfei OTHER seem right. Thanks.

capfei (approved):

PR #22202, opened 2022-11-01
NHunspell 1.2.5359.26126

LandOfBliss:
**Type:** Missing

**Summary:**
NHunspell 1.2.5359.26126

**Details:**
Readme states:  NHunspell is licensed under: GPL/LGPL/MPL. Free use in commercial applications is permitted according to the LGPL and MPL licenses. Your commercial application can link against the NHunspell DLLs.

**Resolution:**
LGPL-2.0-or-later OR GPL-1.0-or-later) OR MPL-1.0

**Affected definitions**:
- [NHunspell 1.2.5359.26126](https://clearlydefined.io/definitions/nuget/nuget/-/NHunspell/1.2.5359.26126/1.2.5359.26126)

### DIFF
--- a/curations/nuget/nuget/-/NHunspell.yaml
+++ b/curations/nuget/nuget/-/NHunspell.yaml
@@ -5,4 +5,4 @@ coordinates:
 revisions:
   1.2.5359.26126:
     licensed:
-      declared: (LGPL-2.0-or-later OR GPL-1.0-or-later) OR MPL-1.0
+      declared: GPL-2.0-or-later OR LGPL-2.1-or-later OR MPL-1.1
--- a/curations/nuget/nuget/-/NHunspell.yaml
+++ b/curations/nuget/nuget/-/NHunspell.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: NHunspell
+  provider: nuget
+  type: nuget
+revisions:
+  1.2.5359.26126:
+    licensed:
+      declared: (LGPL-2.0-or-later OR GPL-1.0-or-later) OR MPL-1.0


capfei:
@ariel11 Did we ever decide on the default for MPL if no version is specified?

ariel11:
@capfei - this is a really interesting one from a curation perspective.  On the one hand, the MPL-2.0 license was released at the time this package was published and MPL-2.0 is much more common than the older versions.  So, from that perspective, I'd be inclined to use MPL-2.0 as the default.  A separate question, however, is whether the license commentary on the README actually changes this to "OTHER." I wonder if that commentary would be considered as changing the terms of the licenses specified or otherwise turns this into a custom license --  "_NHunspell is licensed under: GPL/LGPL/MPL. Free use in commercial applications is permitted according to the LGPL and MPL licenses. Your commercial application can link against the NHunspell DLLs._"

This may be a good one for Wed.  Also adding @copernicat to this one for possible comment.  Thanks!

copernicat:
Hi @ariel11 , thanks for asking. The only tri-license of these 3 I've ever seen uses MPL-1.1.  The official Hunspell repo contains the MPL-1.1, see project's website http://hunspell.github.io/ and specifically https://github.com/hunspell/hunspell/blob/master/COPYING.MPL. This should be curated as MPL-1.1 OR LGPL-2.1-or-later OR GPL-2.0-or-later, according to the official license readme: (https://github.com/hunspell/hunspell/blob/master/license.hunspell)

As to your question, the comment in the read me is an unsolicited interpretation of the license terms and not new or different license terms ("NHunspell is licensed under: GPL/LGPL/MPL. Free use in commercial applications is permitted according to the LGPL and MPL licenses. Your commercial application can link against the NHunspell DLLs.")

PR #22265, opened 2022-11-07
Update curations/gem/rubygems/-/unicorn.yaml

RomanIakovlev:
Harvesting has incorrectly added a whole lot of different licenses here, including the NOASSERTION one, which is clearly wrong. According to my understanding, unicorn is licensed under either GPL-2.0-or-later or Ruby license.

I have only added a couple of latest versions to the curation, but it might make sense to make a similar change for all the older versions as well.

### DIFF
--- a/curations/gem/rubygems/-/unicorn.yaml
+++ b/curations/gem/rubygems/-/unicorn.yaml
@@ -5,4 +5,10 @@ coordinates:
 revisions:
   5.4.0:
     licensed:
-      declared: GPL-2.0-or-later OR OTHER
+      declared: GPL-2.0-or-later OR Ruby
+  6.0.0:
+    licensed:
+      declared: GPL-2.0-or-later OR Ruby
+  6.0.0:
+    licensed:
+      declared: GPL-2.0-or-later OR Ruby


capfei:
@RomanIakovlev The reason it is curated with `OTHER` instead of `Ruby` is because this uses Ruby 1.8 specific license. The text has some differences that does not make it a match to what SPDX has listed as "Ruby". Per the SPDX matching guidelines, we would use `OTHER`.
```
You may (against our _preference_) also use the Ruby 1.8 license terms
which we inherited from the original Mongrel project when we forked it:

=== Ruby 1.8-specific terms (if you're not using the GPL)
```
image

If you agree, then I can update the PR. Let me know.

RomanIakovlev:
Thanks for the feedback @capfei. I'd like to better understand the reason for mismatch. Is it because of the preamble paragraph you've quoted, or because of the diff in the license text itself? In other words, if unicorn changes the substantive part of their license to be the Ruby license verbatim (e.g. replace the `executables` with `binaries` and so on), would it result in a match per SPDX guidelines or not?

capfei:
@RomanIakovlev Based on the diff. We follow the SPDX matching guidelines (blue text is omittable, red text is replaceable, etc..). https://spdx.dev/license-list/matching-guidelines/

Here is how SPDX lists the `Ruby` license: https://spdx.org/licenses/Ruby.html. The only text they list as being changeable are the ordered list numberings. 

So, yes, if unicorn changes their license to be a match with what SPDX states is considered `Ruby` (change `executables` to `binaries`, update the text in `2c`, ....) then we would curate it as `Ruby`. Licenses that do not match a SPDX license identifier are curated as `OTHER`.

Does that help?

RomanIakovlev:
Thanks @capfei , that's a really helpful explanation. I'll update the PR to change Ruby to Other.

capfei (approved):

capfei:
Thank you!

PR #22302, opened 2022-11-10
Resolve case of incorrect Source Url

Sandhya060893:
**Type:** Incorrect

**Summary:**
Resolve case of incorrect Source Url

**Details:**
There is an incorrect url provided which does not point to the commit "b67b58c617".
Wheras the repo https://github.com/unpaper/unpaper has the commit id present.
Path : https://github.com/unpaper/unpaper/search?q=b67b58c617&type=commits

**Resolution:**
The component's source url is being curated as the commit id is not present in the present url and is present in the https://github.com/unpaper/unpaper.
Path: https://github.com/unpaper/unpaper/search?q=b67b58c617&type=commits

**Affected definitions**:
- [unpaper b67b58c6179a5acc1e8909fbf39a1a3def2924be](https://clearlydefined.io/definitions/git/github/flameeyes/unpaper/b67b58c6179a5acc1e8909fbf39a1a3def2924be/b67b58c6179a5acc1e8909fbf39a1a3def2924be)

### DIFF
--- a/curations/git/github/flameeyes/unpaper.yaml
+++ b/curations/git/github/flameeyes/unpaper.yaml
@@ -1,0 +1,15 @@
+coordinates:
+  name: unpaper
+  namespace: flameeyes
+  provider: github
+  type: git
+revisions:
+  b67b58c6179a5acc1e8909fbf39a1a3def2924be:
+    described:
+      sourceLocation:
+        name: unpaper
+        namespace: unpaper
+        provider: github
+        revision: b67b58c6179a5acc1e8909fbf39a1a3def2924be
+        type: git
+        url: 'https://github.com/unpaper/unpaper/commit/b67b58c6179a5acc1e8909fbf39a1a3def2924be'


AE49:
@ariel11 @capfei I think Declared License is GPL-2.0-only - https://github.com/Flameeyes/unpaper/blob/b67b58c6179a5acc1e8909fbf39a1a3def2924be/file.c. I'm not finding mention "or later". Thoughts?

Sandhya060893:
Hi Angela,

Actually there is the Source Url curation request sent as well as it is currently pointing to the url -https://github.com/Flameeyes/unpaper
Whereas the right one seems to be - https://github.com/unpaper/unpaper.

So under the above right url, the LICENSES file has the license mentioned as GPL 2.0 or any later version.
" This program is free software; you can redistribute it and/or modify it under
the terms of the GNU General Public License as published by the Free Software
Foundation; either version 2 of the License, or (at your option) any later
version."

Hence it is being curated as GPL 2.0 or later.

Please do let me know in case of any other concerns.

Thank You.

Best Regards,
Sandhya

capfei:
@Sandhya060893 I don't understand this source correction. Why would we update a fork url with the original source url? This component correctly points to https://github.com/Flameeyes/unpaper/tree/b67b58c6179a5acc1e8909fbf39a1a3def2924be, which does exist in this fork. This curation isn't for unpaper/unpaper but the fork Flameeyes/unpaper.

ariel11:
@capfei - I agree.  This particular definition page of ClearlyDefined is for flameeyes/unpaper/tree/b67b58c6179a5acc1e8909fbf39a1a3def2924be, so the name/namespace and commit ID are correct.  I think we can close this PR, yes?

PR #22348, opened 2022-11-14
add license information for test-unit/test-unit versions 3.3.7 and later

elrayle:
## Description

Querying Clearly Defined for Ruby GEM test-unit/test-unit at version 3.5.3 returns a license that includes `NOASSERTION` and does not match the expected license.

### Actual License returned from Clearly Defined

```
(BSD-2-Clause AND NOASSERTION) OR (BSD-2-Clause AND Ruby) OR (NOASSERTION AND Python-2.0 AND Ruby) OR (Python-2.0 AND Ruby)
```

### Expected License

```
(BSD-2-Clause OR Ruby) AND Python-2.0.1
```

## Analysis

Applies to versions 3.3.7 and later...

#### README

The license statement in the README is...

> This software is distributed under either the terms of new Ruby License or BSDL. See the file [COPYING](https://github.com/test-unit/test-unit/blob/master/COPYING).
> 
> Exceptions
> 
> * lib/test/unit/diff.rb is a triple license of (the new Ruby license or BSDL) and PSF license.

#### License Files

Licenses Referenced: (tag:3.5.5)
* Ruby - [test-unit: COPYING](https://github.com/test-unit/test-unit/blob/3.5.5/COPYING), [SPDX: Ruby](https://spdx.org/licenses/Ruby.html)
* BSDL - [test-unit: BSDL](https://github.com/test-unit/test-unit/blob/3.5.5/BSDL), [SPDX: BSD-2-Clause](https://spdx.org/licenses/BSD-2-Clause.html) _NOTE: In most places, they simply refer to this as `BSDL`, but in COPYING, they specifically say `2-clause BSDL`_
* PSF - [test-unit: PSFL](https://github.com/test-unit/test-unit/blob/3.5.5/PSFL), [SPDX: PSF-2.0](https://spdx.org/licenses/PSF-2.0.html) _Applies to a single file `lib/test/unit/diff.r`._

Statement in COPYING:

> You can redistribute it and/or modify it under either the terms of the2-clause BSDL (see the file BSDL), or the conditions below: _remainder of file is the Ruby license text_
> 
> Exceptions
> ----------
> * lib/test/unit/diff.rb: This license and PSF license

Statement in [lib/test/unit/diff.rb](https://github.com/test-unit/test-unit/blob/master/lib/test/unit/diff.rb)

> It is free software, and is distributed under (the new Ruby license or BSDL) and the PSF license.

### Summary

The statements in the various files referencing licensing leads to the expected license...

```
(BSD-2-Clause OR Ruby) AND Python-2.0.1
```

----

CC @kou

### DIFF
--- a/curations/gem/rubygems/-/test-unit.yaml
+++ b/curations/gem/rubygems/-/test-unit.yaml
@@ -1,0 +1,62 @@
+coordinates:
+  name: test-unit
+  provider: rubygems
+  type: gem
+revisions:
+  3.3.7:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.3.8:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.3.9:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.0:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.1:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.2:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.3:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.4:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.5:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.6:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.7:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.8:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.4.9:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.5.0:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.5.1:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.5.2:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.5.3:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.5.4:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1
+  3.5.5:
+    licensed:
+      declared: (BSD-2-Clause OR Ruby) AND Python-2.0.1  
--- a/curations/gem/rubygems/-/test-unit.yaml
+++ b/curations/gem/rubygems/-/test-unit.yaml
@@ -5,58 +5,58 @@ coordinates:
 revisions:
   3.3.7:
     licensed:
-      declared: (BSD-2-Clause OR Ruby) AND Python-2.0
+      declared: BSD-2-Clause OR Ruby
   3.3.8:
     licensed:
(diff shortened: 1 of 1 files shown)


elrayle:
### Reviewer Attention

Rubygems.org reports 90 versions.  There are at least 8 changes to licensing files across those versions.  This PR only adds licensing for the current version and previous versions that use the same licensing.

NOTE: The text of PSFL matches that of [SPDX Python-2.0.1](https://spdx.org/licenses/Python-2.0.1.html).  Specifying this license failed the automated checks.  [SPDX Python-2.0](https://spdx.org/licenses/Python-2.0..html) is almost identical.  It was used instead in the license statement.

This is my first PR to curated-data.  Please let me know if I need to make some adjustments.

capfei:
@elrayle For this component, the declared would be `BSD-2-Clause OR Ruby` since that is what is stated as the overall license. The callout for the single file would be what we would consider as "discovered" since, in this case, it applies to a single file. 

You can curate the single file to the three licenses, but if you are fine with it, I will update the declared.

As for Python-2.0.1, we have an open issue to get our validation tools updated to the current SPDX license list. We use Python-2.0.1, but have to manually update PRs and then force merge to override the validation. You can keep using Python-2.0 since the PR will be updated to Python-2.0.1 when it is reviewed. Cumbersome but it is what we can do until our tooling is updated. I have to do this with other identifiers that have been added after 3.11.

elrayle:
@capfei Thanks for the info.  Feel free to update the PR as needed to make it ready to merge.

I have a  few followup questions to be sure I understand the process and can make helpful contributions in the future.

### Clarifications

Are the following statements true?
* curated-data holds declared licenses only
* the declared license is generally defined by LICENSE or COPYING (possibly with a file extension)
* the declared license does not include file specific exceptions

These statements lead to the declared license for test-unit to be `BSD-2-Clause OR Ruby` based on the contents of COPYING.

### How is the reported license derived?

Is there a place in https://github.com/clearlydefined or the [UI](https://clearlydefined.io/) that would show me the files that were used to generate the license that is returned by the API?  Looking at the repo source, I'm not seeing how the returned license is generated.

Request:
```
GET https://api.clearlydefined.io/definitions/gem/rubygems/-/test-unit/3.5.3?expand=-files
```

Response:
```
"licensed": {
  "declared": "(BSD-2-Clause AND NOASSERTION) OR (BSD-2-Clause AND Ruby) OR (NOASSERTION AND Python-2.0 AND Ruby) OR (Python-2.0 AND Ruby)"
  ...
}
```

elrayle:
@capfei Based on your feedback, I updated the license to `BSD-2-Clause OR Ruby`.

capfei:
@elrayle I'll answer as best as I can! 😀 

> curated-data holds declared licenses only

`curated-data` has curations for source locations, copyright statement fixes, release dates, and facets (but we really don't accept many of those since facets are not fleshed out). We can also do file-level declared licenses and copyright statements. We are not able to curate the discovered license field. 

> the declared license is generally defined by LICENSE or COPYING (possibly with a file extension

I find that to be the case with things that been published within the last couple of years. I do find that is not the case with older stuff or very small projects. License info can be in READMEs, package metadata (pom.xml, package.json, nuspec, etc..), file headers, project sites, doc files... I'm sure there are other examples that are escaping but I would not confine my search to just the LICENSE or COPYING. Some licenses need clarification from file headers. The main ones I come across GPL/LGPL licenses since you need the headers to determine if you are curating `-or-later` or `only`.

> the declared license does not include file specific exceptions

Usually. We find most of those would fall under "discovered" since they are not defined as the overall project license. More like calling out third-party code. Of course, there are always exceptions. An example would be if they do not declare an overall license but something like "lib folder is under MIT and tests are under BSD".  If the component contains just the `lib` and `tests` folder, then we would curate as `BSD-3-Clause AND MIT`.

capfei:
It looks like your query is excluding the files with  `-files` instead of `files`.
https://api.clearlydefined.io/definitions/gem/rubygems/-/test-unit/3.5.3?expand=files

In the UI, you can click on "Raw Data" then "Harvested Data" to see what each scanner found.
https://clearlydefined.io/definitions/gem/rubygems/-/test-unit/3.5.3

If you want to do it through the API, I think https://api.clearlydefined.io/harvest/gem/rubygems/-/test-unit/3.5.3?form=raw is the same. 

I play around on https://api.clearlydefined.io/api-docs since I can easily see the different queries and options (curation, definition, harvest).

Does that help?

capfei (approved):

elrayle:
@capfei Thanks a lot for the information.  That will be very helpful.

PR #22530, opened 2022-11-30

Sandhya060893:
**Type:** Incorrect

**Summary:**

**Details:**
There are multiple GPL versions licensed specified in whereas the license information from source repo shows it is under GPL 2.0 or later and also LGPL 2.1 or later for some of the files and code.
Path : https://github.com/git/git/blob/v2.16.4/COPYING
https://github.com/git/git/blob/v2.16.4/LGPL-2.1

**Resolution:**
The component is being curated as GPL 2.0 or later AND LGPL 2.1 or later as per the information in the source repository.
Path : https://github.com/git/git/blob/v2.16.4/LGPL-2.1
https://github.com/git/git/blob/v2.16.4/COPYING

**Affected definitions**:
- [git a42a58d7b62cc1d6301440e81a83feed9d7c118c](https://clearlydefined.io/definitions/git/github/git/git/a42a58d7b62cc1d6301440e81a83feed9d7c118c/a42a58d7b62cc1d6301440e81a83feed9d7c118c)

### DIFF
--- a/curations/git/github/git/git.yaml
+++ b/curations/git/github/git/git.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   a42a58d7b62cc1d6301440e81a83feed9d7c118c:
     licensed:
-      declared: GPL-2.0-or-later AND LGPL-2.1-or-later
+      declared: GPL-2.0-only
--- a/curations/git/github/git/git.yaml
+++ b/curations/git/github/git/git.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: git
+  namespace: git
+  provider: github
+  type: git
+revisions:
+  a42a58d7b62cc1d6301440e81a83feed9d7c118c:
+    licensed:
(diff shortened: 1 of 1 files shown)


capfei:
@ariel11 I think think should be GPL-2.0-only per the COPYING file. The LGPL license file does state that it applies to the `xdiff` folder and some files in `compat` but that doesn't seem like it should be included in the declared. The file also says that there are other licenses scattered about and would need to check each file for that information. This seems like it should be under discovered along with any other licenses in the repo.

What do you think?

ariel11:
@capfei - yes, I agree - with the way ClearlyDefined defines declared vs. discovered, the arching project license is GPL-2.0-only.  The README states:

Git is an Open Source project covered by the GNU General Public License version 2 (some parts of it are under different licenses, compatible with the GPLv2). It was originally written by Linus Torvalds with help of a group of hackers around the net.

capfei (approved):

PR #22548, opened 2022-12-01

Sandhya060893:
**Type:** Incorrect

**Summary:**

**Details:**
There is GPL AND LGPL AND MPL mentioned whereas the license file of the source repository says either of GPL or LGPL or MPL-1.1 can be chosen as the license.
Path : https://github.com/ckeditor/ckeditor4/blob/master/LICENSE.md

**Resolution:**
The component is being curated as MPL-1.1 instead of GPL AND LGPL AND MPL-1.1 as per the information given in license file of source repo.
Since we choose a lower risk license over high risk one it is being curated wih MPL-1.1.
Path : https://github.com/ckeditor/ckeditor4/blob/master/LICENSE.md

**Affected definitions**:
- [ckeditor-dev 8a12b041718e05eeffb62a931d95995ac72cbe22](https://clearlydefined.io/definitions/git/github/ckeditor/ckeditor-dev/8a12b041718e05eeffb62a931d95995ac72cbe22/8a12b041718e05eeffb62a931d95995ac72cbe22)

### DIFF
--- a/curations/git/github/ckeditor/ckeditor-dev.yaml
+++ b/curations/git/github/ckeditor/ckeditor-dev.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   8a12b041718e05eeffb62a931d95995ac72cbe22:
     licensed:
-      declared: MPL-1.1
+      declared: GPL-2.0-or-later OR LGPL-2.1-or-later OR MPL-1.1
--- a/curations/git/github/ckeditor/ckeditor-dev.yaml
+++ b/curations/git/github/ckeditor/ckeditor-dev.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: ckeditor-dev
+  namespace: ckeditor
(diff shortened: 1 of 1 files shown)


capfei:
I think this should be `GPL-2.0-or-later OR LGPL-2.1.-or-later OR MPL-1.1` since the license states you can choose either of these to use as the main license.

@AE49 Do you agree?

AE49:
@capfei I agree with the first two licenses in your curation. I see "MPL-1.1.-or-later" though here -https://github.com/ckeditor/ckeditor4/blob/8a12b041718e05eeffb62a931d95995ac72cbe22/LICENSE.md#software-license-agreement

capfei:
@AE49 Since there isn't `MPL-1.1-or-later`, what should we use?

AE49:
@capfei I get it now...I agree with your curation.

capfei (approved):

PR #23319, opened 2023-01-31
paramiko@2.11.0 LGPL-2.1-or-later curation

dangoor:
The license file shows that it's LGPL (and there are other curations that effect).


dangoor:
I'm going to update this PR to use the existing Paramiko curation file.

capfei (approved):

PR #23363, opened 2023-02-02
LGPL-3.0-or-later

LandOfBliss:
**Type:** Missing

**Summary:**
LGPL-3.0-or-later

**Details:**
POM indicates GNU LGPLv3 and links to license: https://www.gnu.org/licenses/lgpl-3.0.txt

which is LGPL-3.0-or-later

**Resolution:**
LGPL-3.0-or-later

**Affected definitions**:
- [jffi 1.0.1](https://clearlydefined.io/definitions/maven/mavencentral/org.jruby.extras/jffi/1.0.1/1.0.1)

### DIFF
--- a/curations/maven/mavencentral/org.jruby.extras/jffi.yaml
+++ b/curations/maven/mavencentral/org.jruby.extras/jffi.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: jffi
+  namespace: org.jruby.extras
+  provider: mavencentral
+  type: maven
+revisions:
+  1.0.1:
+    licensed:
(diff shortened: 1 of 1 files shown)


capfei:
I can't anything that indicates "or later" since there is no source zip and the source repo is gone. With that, I think this should be "LGPL-3.0-only". @LandOfBliss Do you agree with this?

LandOfBliss:
> I can't anything that indicates "or later" since there is no source zip and the source repo is gone. With that, I think this should be "LGPL-3.0-only". @LandOfBliss Do you agree with this?

Yes, I agree

capfei (approved):

PR #23368, opened 2023-02-02
org.jruby.extras/jnr-netdb0.4

LandOfBliss:
**Type:** Missing

**Summary:**
org.jruby.extras/jnr-netdb0.4

**Details:**
Maven pom indicates: 
LGPL-3.0-or-later

https://repo1.maven.org/maven2/org/jruby/extras/jnr-netdb/0.4/jnr-netdb-0.4.pom

GIthub link broken

**Resolution:**
LGPL-3.0-or-later

**Affected definitions**:
- [jnr-netdb 0.4](https://clearlydefined.io/definitions/maven/mavencentral/org.jruby.extras/jnr-netdb/0.4/0.4)

### DIFF
--- a/curations/maven/mavencentral/org.jruby.extras/jnr-netdb.yaml
+++ b/curations/maven/mavencentral/org.jruby.extras/jnr-netdb.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   '0.4':
     licensed:
-      declared: LGPL-3.0-or-later
+      declared: LGPL-3.0-only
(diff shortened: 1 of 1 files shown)


capfei:
Since there isn't anything in the package that states "or later", no source zip, and the source repo is gone I think this should be LGPL-3.0-only. @LandOfBliss Do you agree?

LandOfBliss:
> Since there isn't anything in the package that states "or later", no source zip, and the source repo is gone I think this should be LGPL-3.0-only. @LandOfBliss Do you agree?

Yes, I agree

capfei (approved):

PR #23598, opened 2023-02-22
Add license curation for github.com/hhatto/gorst@v0.0.0-2018102913320…

dangoor:
Summary:
github.com/hhatto/gorst v0.0.0-20181029133204-ca9f730cac5b curation

Details:
github.com/hhatto/gorst is under BSD-3-Clause AND (GPL-2.0-or-later OR MIT) license for the v0.0.0-20181029133204-ca9f730cac5b version
Resolution:
https://github.com/hhatto/gorst/blob/ca9f730cac5b/LICENSE (as does the LICENSE on the master branch) says:
> peg-markdown is released under both the GPL and MIT licenses.
You may pick the license that best fits your needs.

### DIFF
--- a/curations/go/golang/github.com%2Fhhatto/gorst.yaml
+++ b/curations/go/golang/github.com%2Fhhatto/gorst.yaml
@@ -6,4 +6,9 @@ coordinates:
 revisions:
   v0.0.0-20181029133204-ca9f730cac5b:
     licensed:
-      declared: BSD-3-Clause AND (GPL-2.0-or-later OR MIT)
+      declared: GPL-2.0-or-later OR MIT
+    facets:
+      core:
+        discovered:
+          expressions:
+            - BSD-3-Clause
--- a/curations/go/golang/github.com%2Fhhatto/gorst.yaml
+++ b/curations/go/golang/github.com%2Fhhatto/gorst.yaml
@@ -7,4 +7,4 @@
     v0.0.0-20181029133204-ca9f730cac5b:
       licensed:
         declared:
-          - BSD-3-Clause
+          - BSD-3-Clause AND (GPL-2.0-or-later OR MIT)
--- a/curations/go/golang/github.com%2Fhhatto/gorst.yaml
+++ b/curations/go/golang/github.com%2Fhhatto/gorst.yaml
@@ -1,10 +1,9 @@
-- coordinates:
-    name: github.com/hhatto/gorst
-    namespace: github.com%2Fhhatto
-    provider: golang
-    type: go
-  revisions:
-    v0.0.0-20181029133204-ca9f730cac5b:
-      licensed:
-        declared:
-          - BSD-3-Clause AND (GPL-2.0-or-later OR MIT)
+coordinates:
+  name: github.com/hhatto/gorst
+  namespace: github.com%2Fhhatto
+  provider: golang
+  type: go
+revisions:
+  v0.0.0-20181029133204-ca9f730cac5b:
+    licensed:
+      declared: BSD-3-Clause AND (GPL-2.0-or-later OR MIT)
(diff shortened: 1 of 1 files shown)


capfei:
@dangoor I'm not familiar with Go so let me know if I am mistaken. 

Looking around, other than the Makefile that points to the repo (github.com/knieriem/peg) that uses BSD-3-Clause, I'm not able to find anything else with BSD-3-Clause. The file headers either having no copyright info or state:
```
/*  Original C version https://github.com/jgm/peg-markdown/
 *	Copyright 2008 John MacFarlane (jgm at berkeley dot edu).
 *
 *  Modifications and translation from C into Go
 *  based on markdown_lib.c and parsing_functions.c
 *	Copyright 2010 Michael Teichgräber (mt at wmipf dot de)
 *
 *  This program is free software; you can redistribute it and/or modify
 *  it under the terms of the GNU General Public License or the MIT
 *  license.  See LICENSE for details.
 *
 *  This program is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU General Public License for more details.
 */
```

To me, this reads like the declared should be `GPL-2.0-or-later OR MIT`. Do you think this should be updated?

dangoor:
Oh, you're right. This is more confusing than it seemed at first. If you look at the [license file](https://github.com/hhatto/gorst/blob/ca9f730cac5b/LICENSE) in the package, you can see the BSD-3-Clause license at the end (with a copyright notice crediting the Go authors). Maybe they added that because of the use of the `goimports` tool. I assume the LICENSE file is where the scanner found that license.

Seems like a bit of a mishmash of code from different sources. I think the presence of the BSD-3-Clause license in the LICENSE file argues for my proposed declared value.

capfei:
@ariel11 @AE49 Can I get your thoughts? I'm still conflicted and leaning towards BSD-3-Clause as "Discovered" instead of "Declared".

dangoor:
I'll note that my primary concern with this curation is the `GPL-2.0-or-later OR MIT` part and I don't really have a solid opinion on the `BSD-3-Clause`, since it is quite difficult to tell why that's even there.

ariel11:
@capfei - yes, agree based on how we define "declared" and "discovered" license info in the[ curation guidelines](https://github.com/clearlydefined/clearlydefined/blob/master/docs/curation-guidelines.md), that BSD-3-Clause would be classified as a "discovered" license.  Agree that "GPL-2.0-or-later OR MIT" appears to be the original license info that flows through to this code.

dangoor:
I've updated the curation to move the BSD-3-Clause to discovered. I _think_ I did it correctly (based on where discovered lives in the output from ClearlyDefined).

capfei:
@dangoor I have never done a curation on that so we'll both see if it works! 😆 If it doesn't, I will remove it.

capfei (approved):

capfei:
@dangoor Just to follow up. The facets info didn't work. We were always told we couldn't updated the "Discovered" field so this further confirms that we are not able to. 😞 

Another thing I didn't notice yesterday were some of the errors in the curation that caused the ClearlyDefined site to not update with the new license info.
https://clearlydefined.io/definitions/go/golang/github.com%2Fhhatto/gorst/v0.0.0-20181029133204-ca9f730cac5b

First, the file saved to the wrong path: `curations/go/golang/github.com%2Fhhatto/gorst.yaml`
I had to move it to `curations/go/golang/github.com/hhatto/gorst.yaml`.

Also, the `name` value was wrong. It should have been `gorst` instead of `github.com/hhatto/gorst`.
Here is the commit with the name update and relocation: https://github.com/clearlydefined/curated-data/commit/307cad9c07d1f7ab05393dca9ceb71e787ff8d18

dangoor:
@capfei Good to know about the "Discovered" field.

I'll work on fixing those errors you found with our automation. Sorry for the extra pain there!

PR #23649, opened 2023-02-28
Fixing file and deleting incorrect location

capfei:
Didn't catch these the first time:
- `name` is incorrect; should have been `gorst`
- file location is not in correct location


capfei:
I'm assuming the error is from the facets part but the error message doesn't state that. Will try merging to see if the discovered does update. If not, I'll fix it.

PR #23651, opened 2023-02-28
Add license curation for chardet@5.1.0

dangoor:
Summary:
chardet 5.1.0 curation

Details:
chardet is under LGPL-2.1-only for the 5.1.0 version
Resolution:
LGPL license here: https://github.com/chardet/chardet/blob/8087f0015184cf9569d00970a2acf1e1d3e80feb/LICENSE

### DIFF
--- a/curations/pypi/pypi/-/chardet.yaml
+++ b/curations/pypi/pypi/-/chardet.yaml
@@ -9,3 +9,6 @@ revisions:
   4.0.0:
     licensed:
       declared: LGPL-2.1-only
+  5.1.0:
+    licensed:
+      declared: LGPL-2.1-only


capfei:
@dangoor Looking at the file headers, it states:
```
######################## BEGIN LICENSE BLOCK ########################
# The Original Code is Mozilla Universal charset detector code.
#
# The Initial Developer of the Original Code is
# Netscape Communications Corporation.
# Portions created by the Initial Developer are Copyright (C) 2001
# the Initial Developer. All Rights Reserved.
#
# Contributor(s):
#   Mark Pilgrim - port to Python
#   Shy Shalom - original C code
#
# This library is free software; you can redistribute it and/or
# modify it under the terms of the GNU Lesser General Public
# License as published by the Free Software Foundation; either
# version 2.1 of the License, or (at your option) any later version.
#
# This library is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
# Lesser General Public License for more details.
#
# You should have received a copy of the GNU Lesser General Public
# License along with this library; if not, write to the Free Software
# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
# 02110-1301  USA
######################### END LICENSE BLOCK #########################
```

I think this should be `LGPL-2.1.-or-later`. If you agree, I can update the PR. Let me know.

dangoor:
@capfei You're right. I had only looked at the LICENSE file and not a file header. `LGPL-2.1-or-later` makes sense to me. I would bet that license is unchanged from the prior version that's also in the curation file. I don't know if it's worth changing both. Thanks for spotting that!

capfei:
@dangoor Yep, it looks like 3.0.4 and 4.0.0 need to corrected to LGPL-2.1-or-later. It's not letting me make changes to the PR so I will merge this so it adds 5.1.0 to the file and then update all 3 versions at once. 😄

PR #23699, opened 2023-03-01
hotchocolate.types.scalars.upload MIT License

clearlydefinedbot:
**Type:** Incomplete

**Summary:**
hotchocolate.types.scalars.upload MIT License

**Details:**
Add MIT license per Nuget and GitHub Repo

**Resolution:**
Declare MIT license

**Affected definitions**:
- [hotchocolate.types.scalars.upload 12.18.0](https://clearlydefined.io/definitions/nuget/nuget/-/hotchocolate.types.scalars.upload/12.18.0/12.18.0)

PR #23700, opened 2023-03-01
Delete hotchocolate.types.scalars.upload.yaml

capfei:
Incorrect casing. Deleting file since file with correct casing already exists.


capfei:
Fixes #23699

PR #24063, opened 2023-05-04
Declared License is missing

SwadhinMangaraj:
**Type:** Missing

**Summary:**
Declared License is missing

**Details:**
Declared license was missing.

**Resolution:**
Updated the declared license as per https://github.com/scikit-image/scikit-image/blob/v0.20.0/LICENSE.txt.

**Affected definitions**:
- [scikit-image 5e74a4a3a5149a8a14566b81a32bb15499aa3857](https://clearlydefined.io/definitions/git/github/scikit-image/scikit-image/5e74a4a3a5149a8a14566b81a32bb15499aa3857/5e74a4a3a5149a8a14566b81a32bb15499aa3857)

### DIFF
--- a/curations/git/github/scikit-image/scikit-image.yaml
+++ b/curations/git/github/scikit-image/scikit-image.yaml
@@ -6,4 +6,4 @@ coordinates:
 revisions:
   5e74a4a3a5149a8a14566b81a32bb15499aa3857:
     licensed:
-      declared: BSD-3-Clause AND BSD-2-Clause AND MIT AND OTHER
+      declared: BSD-3-Clause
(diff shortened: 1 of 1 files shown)


capfei:
@ariel11 Do we consider all of the individual callouts as part of the overall license since this is for the GitHub repo or look at them as third party code?

ariel11:
@capfei - Well, I think when the license file makes a distinction between a default/overall license for most of the project and then licenses that apply only to a subset of files, we use that to distinguish declared vs discovered in a repo.  Here, I read the license file as saying the 'main' license for all the files, unless otherwise called out, is BSD-3-Clause.  And then the other licenses apply to a small set of files.  So my recommendation on this would be BSD-3-Clause for declared and the rest as discovered. Thoughts?

capfei (approved):

PR #24616, opened 2023-07-03
Revert "org.glassfish.hk2:hk2-locator 2.2.0-b21"

AE49:


AE49:
Tried to fix curation but version is not offered in this PR.

PR #25983, opened 2023-10-30
Curation for PyPI Pillow package

dangoor:
As seen from all of the previous curations, Pillow is licensed HPND.

See also [the license file](https://github.com/python-pillow/Pillow/blob/main/LICENSE).


capfei:
@dangoor It looks like this PR has more than a curation in it. Did you mean to include a GitHub action, README update and updating the package.json?

dangoor:
🤦 That's what I get for doing this manually (we were seeing issues with tooling not being able to open PRs for some reason). I'll fix this.

dangoor:
OK, fixed!

capfei:
@dangoor Thanks for the update!

I have had this discussion about [other Pillow curations](https://github.com/clearlydefined/curated-data/pull/21167) but I think this should be updated to MIT-CMU. Following the SPDX matching guidelines, this is a match. Also, SPDX uses Pillow as a reference to this license.
https://spdx.org/licenses/MIT-CMU.html
image

What do you think?

dangoor:
@capfei Interesting that it's a text match for MIT-CMU. In fact, it looks like the text is a match for MIT-CMU but [not for HPND](https://spdx.org/licenses/HPND.html) despite the fact that [Pillow's license file says it's HPND](https://github.com/python-pillow/Pillow/blob/main/LICENSE)

I think you're right that the more correct license is MIT-CMU.

AE49:
@ariel11 for visibility on this because the license says "HPND", however, as @capfei pointed out the SPDX match is MIT-CMU.

dangoor:
It may be worth opening a pull request with the Pillow project to change the license file to remove the HPND sentence, since it's incorrect.

ariel11:
@capfei - what should we do here? Technically, with the colored font on the SPDX templates, one could determine the pillow license ([this portion](https://github.com/python-pillow/Pillow/blob/main/LICENSE#L16-L30)) matches the HPND or MIT-CMU SPDX identifiers. I agree the MIT-CMU is the more complete SPDX identifier but then the pillow license calls itself HPND. Thoughts on opening an Issue or PR on this with pillow?

capfei:
Sorry for the long delay here. I have opened an issue with Pillow asking about either changing the license name to MIT-CMU or simply listing the SPDX identifer in the license file.
https://github.com/python-pillow/Pillow/issues/7942

HPND does not include the first paragraph in the matchable text so I would argue that it is not a match at all and consider it to be a part of the license. 
image

capfei:
@dangoor Pillow has updated their license identifier to MIT-CMU! 
https://github.com/python-pillow/Pillow/issues/7942
https://github.com/python-pillow/Pillow/pull/8460/files

It's almost a year later. Do you want to update your PR or I can add these versions when I update the rest of the curated versions to MIT-CMU. Let me know.

dangoor:
@capfei I went ahead and updated this PR to include the new license. Feel free to either merge this one directly, or if you have a different approach in mind just go ahead and add those new versions in your curation change.

Thanks so much for following up on this!

capfei (approved):

PR #26134, opened 2023-11-21
update go-vcr license for v2.1.0

zkoppert:
Here is the license text for v2.1.0, https://github.com/dnaeon/go-vcr/blob/v2.0.1/LICENSE

Hasn't changed from previous versions but this version is missing from the curation list.

### DIFF
--- a/curations/go/golang/github.com/dnaeon/go-vcr.yaml
+++ b/curations/go/golang/github.com/dnaeon/go-vcr.yaml
@@ -22,3 +22,6 @@ revisions:
   v1.2.0:
     licensed:
       declared: BSD-2-Clause
+  v2.0.1:
+    licensed:
+      declared: BSD-2-Clause


capfei:
@zkoppert Please let me know if I am misunderstanding, but it looks like version 2.0.1 is for [github.com/dnaeon/go-vcr/v2](https://pkg.go.dev/github.com/dnaeon/go-vcr/v2@v2.0.1) instead of [github.com/dnaeon/go-vcr](https://pkg.go.dev/github.com/dnaeon/go-vcr). The path that would need to be curated would be `/go/golang/github.com/dnaeon/go-vcr/v2`.

Also, if it is v2, it is already declared as BSD-2-Clause and this PR can be closed.
https://clearlydefined.io/definitions/go/golang/github.com%2fdnaeon%2fgo-vcr/v2/v2.0.1

PR #26155, opened 2023-11-21
Create vitess.yaml

zkoppert:
License for v0.12.3 at https://github.com/vitessio/vitess/blob/v0.12.3/LICENSE

### DIFF
--- a/curations/go/golang/github.com/vitess.io/vitess.yaml
+++ b/curations/go/golang/github.com/vitess.io/vitess.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: vitess
+  namespace: github.com%2vitessio
+  provider: golang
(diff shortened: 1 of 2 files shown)


zkoppert:
Not sure if the file path for this one should be `vitess.io/vitess.yaml` or `vitessio/vitess.yaml`

capfei:
@zkoppert It looks like the namespace does not include github.com so it should just be vitess.io.
` namespace: vitess.io`
https://pkg.go.dev/vitess.io/vitess@v0.12.3


Also, the file path should be `vitess.io/vitess.yaml`

zkoppert:
@capfei Thanks! Got that corrected.

capfei (approved):

PR #26166, opened 2023-11-28
Add version 2.4.2-pre.22 to xunit

zkoppert:
Add another build for which the license information has not changed.


capfei:
@zkoppert NuGet does not have a version `2.4.2-pre.22` for `xunit`. Is this the right version number?
image

zkoppert:
Interesting! It's called out in one of our csproj files. Let me do some digging.

PR #26193, opened 2023-12-06
Add vscode-tas-client license

lszomoru:


capfei:
@lszomoru Thank you for the curation! It looks like when the PR was submitted this version had not been harvested. I initiated a harvest and it finally updated today. It came back with the declared as MIT so I will close this PR since it is no longer needed.
https://clearlydefined.io/definitions/npm/npmjs/-/vscode-tas-client/0.1.75

PR #26194, opened 2023-12-06
Add tas-client license

lszomoru:


capfei:
Similar to #26193; the version had not been harvested. The harvest finally updated today and came back with the declared as MIT. I will close this PR since it is no longer needed. Thank you!
https://clearlydefined.io/definitions/npm/npmjs/-/tas-client/0.2.33

PR #26657, opened 2024-02-15
Add curation for go cronexpr

dangoor:
The correct license is Apache-2.0 OR GPL-3.0-only which can be found at https://github.com/hashicorp/cronexpr/blob/master/README.md#license

### DIFF
--- a/curations/go/golang/github.com/hashicorp/cronexpr.yaml
+++ b/curations/go/golang/github.com/hashicorp/cronexpr.yaml
@@ -1,0 +1,9 @@
+coordinates:
+    type: go
+    provider: golang
+    namespace: github.com/hashicorp
+    name: cronexpr
+revisions:
+    1.1.1:
+        licensed:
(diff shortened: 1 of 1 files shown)


capfei (approved):

capfei:
@dangoor I missed some small changes needed before merging! 🤦‍♀️ The namespace should have been `github.com%2Fhashicorp`. This is something with ClearlyDefined and handling of the additional slashes in go package names. The other is the version should be `v1.1.1`. This is how it is listed in the package.
https://pkg.go.dev/github.com/hashicorp/cronexpr@v1.1.1

I updated the file with these changes. Sorry I missed them before merging! 
https://github.com/clearlydefined/curated-data/blob/master/curations/go/golang/github.com/hashicorp/cronexpr.yaml

dangoor:
@capfei Oh, thanks for pointing those out! I'm kind of surprised that our tooling got those wrong, but I'll make sure it gets fixed.

PR #27036, opened 2024-04-04
Add curation for pypi Pillow

theia-jane:
The correct license is HPND License which can be found at https://github.com/python-pillow/Pillow/blob/main/LICENSE


capfei:
Raised question with Pillow since license text match SPDX identifier MIT-CMU, not HPND since it does not include the first paragraph. Waiting for outcome on https://github.com/python-pillow/Pillow/issues/7942.

PR #27346, opened 2024-05-07
ci: fix build script to ignore directories with yaml in the name

ralphbean:
Currently in the repo, there are two directories that end with .yaml which then erroneously appear to the build script as if they were yaml files.

* curations/maven/mavencentral/org.yaml
* curations/sourcearchive/mavencentral/org.yaml

This change adjusts the build script to only process paths which match the glob and which are actual files, not directories.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -17,7 +17,7 @@ const colors = {
 
 // @todo only load new/changed yaml - use git diff
 function getYamlFilePaths() {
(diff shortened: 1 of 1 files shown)


ralphbean:
Is this script even used, though? (It does not appear to be used by the clearlydefined check result.)

lumaxis:
As far as I can tell, this is not used in CI but there is an npm script for it and a VS Code task. I think having this fixed is better than not, thanks for helping out!

PR #28445, opened 2024-08-30
chore: adds additional version license information for std uri template npm package

baywet:
follow up of #28444

### DIFF
--- a/curations/npm/npmjs/@std-uritemplate/std-uritemplate.yaml
+++ b/curations/npm/npmjs/@std-uritemplate/std-uritemplate.yaml
@@ -4,6 +4,24 @@ coordinates:
   provider: npmjs
   type: npm
 revisions:
+  1.0.0:
+    licensed:
+      declared: Apache-2.0
+  1.0.1:
+    licensed:
+      declared: Apache-2.0
   1.0.2:
+    licensed:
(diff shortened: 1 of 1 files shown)


capfei:
Some versions haven't been harvested yet so I will wait for those to finish before merging.

baywet:
if the only version that's missing is 1.0.6, I'm happy to remove it from this PR as [I PRed the dependency directly to fix the license information](https://github.com/std-uritemplate/std-uritemplate/pull/233) with the hope that we won't need to manually fix the license information moving forward.

How do I know whether a given version has been harvested? Is my assumption with the license fix correct?

curations/npm/npmjs/@std-uritemplate/std-uritemplate.yaml
```diff
@@ -4,6 +4,24 @@ coordinates:
...
+      declared: Apache-2.0
+  1.0.4:
+    licensed:
+      declared: Apache-2.0
+  1.0.5:
+    licensed:
+      declared: Apache-2.0
+  1.0.6:
+    licensed:
+      declared: Apache-2.0
```
baywet, line 27:
```suggestion
```

capfei (approved):

PR #28669, opened 2024-10-02
Add curation for go tools-golang

Ahmed3lmallah:
The correct license is (Apache-2.0 OR GPL-2.0-or-later) OR CC-BY-4.0 which can be found at https://github.com/spdx/tools-golang/blob/v0.5.4/LICENSE.code and https://github.com/spdx/tools-golang/blob/v0.5.4/LICENSE.docs

### DIFF
--- a/curations/go/golang/github.com/spdx/tools-golang.yaml
+++ b/curations/go/golang/github.com/spdx/tools-golang.yaml
@@ -1,12 +1,12 @@
 coordinates:
-    type: go
-    provider: golang
-    namespace: github.com%2Fspdx
-    name: tools-golang
+  type: go
+  provider: golang
+  namespace: github.com%2Fspdx
+  name: tools-golang
 revisions:
-    v0.5.3:
-        licensed:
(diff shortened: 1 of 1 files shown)


capfei:
@Ahmed3lmallah I think it should be `AND CC-BY-4.0` since that is the license for the documentation, not the code. If you agree, please update the PR. Also, can you use 2 spaces instead of 4? Everything else in this repo uses 2 and it makes for easier reading and keeps consistency.

Ahmed3lmallah:
> @Ahmed3lmallah I think it should be `AND CC-BY-4.0` since that is the license for the documentation, not the code. If you agree, please update the PR. Also, can you use 2 spaces instead of 4? Everything else in this repo uses 2 and it makes for easier reading and keeps consistency.

@capfei  Thank you for reviewing the PR. I have updated the PR and fixed indentation.

capfei (approved):

capfei:

PR #29001, opened 2024-11-10
Add curation for go clock

davecheney:
The correct license is It's Apache-2.0 which can be found at https://github.com/cloudfoundry/clock/blob/main/LICENSE


davecheney:
Sorry, our automation went amock

PR #29106, opened 2024-12-02
Add curation for gem rubocop

bensheldon:
at https://github.com/rubocop/rubocop/blob/master/LICENSE.txt

I think that there is a reference to a Creative Commons license that is specifically [limited to the logo](https://github.com/rubocop/rubocop/tree/master?tab=readme-ov-file#logo):

> The logo is licensed under a [Creative Commons Attribution-NonCommercial 4.0 International License](https://creativecommons.org/licenses/by-nc/4.0/deed.en_GB).

### DIFF
--- a/curations/gem/rubygems/-/rubocop.yaml
+++ b/curations/gem/rubygems/-/rubocop.yaml
@@ -1,11 +1,15 @@
 coordinates:
-  name: rubocop
-  provider: rubygems
-  type: gem
+    type: gem
+    provider: rubygems
+    namespace: ""
+    name: rubocop
 revisions:
-  1.38.0:
-    licensed:
-      declared: MIT
-  1.39.0:
-    licensed:
-      declared: MIT
+    1.38.0:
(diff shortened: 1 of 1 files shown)


curations/gem/rubygems/-/rubocop.yaml
```diff
@@ -1,11 +1,15 @@
 coordinates:
-  name: rubocop
-  provider: rubygems
-  type: gem
+    type: gem
+    provider: rubygems
+    namespace: ""
```
capfei, line 4: I'm not sure if having `""` as the namespace will cause issues since when there isn't one it usually is omitted or "-". Asking in Discord.

capfei:
From my understanding, it shouldn't cause an issue but I will remove it after merging to keep it consistent with the other curations.

capfei:
Also, I will update the spacing to 2 to keep the formatting consistent.

capfei (approved):

PR #29240, opened 2025-01-27
Add BouncyCastle JDK18on license as MIT

dreisch-coveo:
I largely copied the existing `bcprov-jdk15.yaml` file.

**License**
As per https://www.bouncycastle.org/licence.html

image

I think we will save everyone a lot of time and effort by marking this as a license already known to people and tooling.

**Version**
Starts at 1.70 on maven.

image
https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk18o

### DIFF
--- a/curations/maven/mavencentral/bouncycastle/bcprov-jdk18.yaml
+++ b/curations/maven/mavencentral/bouncycastle/bcprov-jdk18.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: bcprov-jdk18on
+  namespace: bouncycastle
+  provider: mavencentral
+  type: maven
+revisions:
+  '1.70':
+    licensed:
+      declared: MIT


capfei:
@dreisch-coveo Thank you for the curation! I agree that Bouncy Castle uses MIT, but the information provided looks incorrect. The package is `org.bouncycastle:bcprov-jdk18on`, not `bouncycastle:bcprov-jdk18on`. Also, I don't see a version `1.70`. It starts at `1.71`, which has already been curated. It looks like version `1.80` needs correcting.
https://clearlydefined.io/definitions/maven/mavencentral/org.bouncycastle/bcprov-jdk18on/1.80

YAML file location: https://github.com/clearlydefined/curated-data/blob/master/curations/maven/mavencentral/org.bouncycastle/bcprov-jdk18on.yaml

dreisch-coveo:
> @dreisch-coveo Thank you for the curation!

Hey, thanks for the feedback. I figured out how to do this via the UI instead, producing this, more correct I suppose, PR: https://github.com/clearlydefined/curated-data/pull/29252

PR #29248, opened 2025-01-27
Add curation for npm applicationinsights-core-js

devm33:
at https://github.com/microsoft/ApplicationInsights-JS/blob/main/shared/AppInsightsCore/LICENSE

### DIFF
--- a/curations/npm/npmjs/@microsoft/applicationinsights-core-js.yaml
+++ b/curations/npm/npmjs/@microsoft/applicationinsights-core-js.yaml
@@ -1,0 +1,9 @@
+coordinates:
+    type: npm
+    provider: npmjs
+    namespace: '@microsoft'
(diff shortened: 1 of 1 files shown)


capfei:
@devm33 Thank you for the contribution! Can you update the spacing to be 2 instead of 4? We use 2. Same for #29249 and #29250.

If not, that's ok. I can fix it after merging. Thank you!

devm33:
> @devm33 Thank you for the contribution! Can you update the spacing to be 2 instead of 4? We use 2. Same for #29249 and #29250.
> 
> If not, that's ok. I can fix it after merging. Thank you!

Done. FWIW these PRs were created with tooling from GitHub, so maybe worth tracking that down to avoid the issue

capfei (approved):

PR #29254, opened 2025-01-28
Add curation for pypi jedi

joshrowley:
at https://github.com/davidhalter/jedi/blob/master/LICENSE.txt


capfei:
@joshrowley Thank you for the contribution! Can you update the spacing to be 2 instead of 4? We use 2. If not, that's ok. I can fix it after merging. Thank you!

capfei (approved):

capfei:
I'll fix after merging.

joshrowley:
Thank you ❤️ , sorry I didn't follow up right away, I appreciate it.

PR #29322, opened 2025-02-04
Add curation for npm unicode-property-aliases-ecmascript

juanrgon:
at https://github.com/mathiasbynens/unicode-property-aliases-ecmascript/blob/main/LICENSE-MIT.txt


capfei:
I'll fix the spacing after merging.

capfei (approved):

PR #29370, opened 2025-02-18
Add curation for npm applicationinsights-core-js

devm33:
at https://github.com/microsoft/ApplicationInsights-JS/blob/main/shared/AppInsightsCore/LICENSE


capfei:
Duplicate of #29372

PR #29374, opened 2025-02-18
Add curation for npm applicationinsights-channel-js

devm33:
at https://github.com/microsoft/ApplicationInsights-JS/blob/main/channels/applicationinsights-channel-js/LICENSE

### DIFF
--- a/curations/npm/npmjs/@microsoft/applicationinsights-channel-js.yaml
+++ b/curations/npm/npmjs/@microsoft/applicationinsights-channel-js.yaml
@@ -1,12 +1,12 @@
 coordinates:
-    type: npm
-    provider: npmjs
-    namespace: '@microsoft'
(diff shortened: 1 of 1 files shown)


curations/npm/npmjs/@microsoft/applicationinsights-channel-js.yaml
```diff
@@ -7,3 +7,6 @@ revisions:
   3.3.4:
     licensed:
       declared: MIT
+  3.3.5:
```
devm33, line 10: Is it possible to set wider version ranges?

capfei: Are you referring to curating a range instead of a version number? If so, that is a requested feature but I don't think there is any planned work on it.

capfei (approved):

PR #29439, opened 2025-03-03
Curate de.siegmar:logback-gelf

kaushalaroraharman:
Curate de.siegmar:logback-gelf with license information

### DIFF
--- a/curations/maven/mavencentral/de.siegmar/logback-gelf.yaml
+++ b/curations/maven/mavencentral/de.siegmar/logback-gelf.yaml
@@ -1,0 +1,9 @@
+coordinates:
+  name: logback-gelf
+  namespace: de.siegmar
+  provider: mavencentral
+  type: maven
+revisions:
+  6.0.1:
+    licensed:
+      declared: LGPL-2.1-only


capfei:
I think this should be `LGPL-2.1-or-later` based on the source file headers. `only` and `or-later` use the same license text, so, typically, you will need to check file headers to verify which applies per SPDX: [only](https://spdx.org/licenses/LGPL-2.1-only.html) or [or-later](https://spdx.org/licenses/LGPL-2.1-or-later.html). In this case, it looks to be `or-later` per the file headers.

https://github.com/osiegmar/logback-gelf/blob/v6.0.1/src/main/java/de/siegmar/logbackgelf/AbstractGelfAppender.java#L5-L8
```
 * Logback GELF - zero dependencies Logback GELF appender library.
 * Copyright (C) 2018 Oliver Siegmar
 *
 * This library is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License as published by the Free Software Foundation; either
 * version 2.1 of the License, or (at your option) any later version.
```

If you agree, can you update the PR?

kaushalaroraharman:
> I think this should be `LGPL-2.1-or-later` based on the source file headers. `only` and `or-later` use the same license text, so, typically, you will need to check file headers to verify which applies per SPDX: [only](https://spdx.org/licenses/LGPL-2.1-only.html) or [or-later](https://spdx.org/licenses/LGPL-2.1-or-later.html). In this case, it looks to be `or-later` per the file headers.
> 
> https://github.com/osiegmar/logback-gelf/blob/v6.0.1/src/main/java/de/siegmar/logbackgelf/AbstractGelfAppender.java#L5-L8
> 
> ```
>  * Logback GELF - zero dependencies Logback GELF appender library.
>  * Copyright (C) 2018 Oliver Siegmar
>  *
>  * This library is free software; you can redistribute it and/or
>  * modify it under the terms of the GNU Lesser General Public
>  * License as published by the Free Software Foundation; either
>  * version 2.1 of the License, or (at your option) any later version.
> ```
> 
> If you agree, can you update the PR?

I have updated the PR with declared license as LGPL-2.1-or-later

capfei:
@kaushalaroraharman Thank you for your contribution!

capfei (approved):

PR #29574, opened 2025-03-25
Add curation for gem diff-lcs

btanasie:
The correct license is (GPL-2.0-or-later OR Artistic-2.0 OR MIT) which can be found at https://github.com/halostatue/diff-lcs/blob/bb28cd9c8e395cabb3a905cb651249133169b2e6/LICENCE.md

### DIFF
--- a/curations/gem/rubygems/-/diff-lcs.yaml
+++ b/curations/gem/rubygems/-/diff-lcs.yaml
@@ -1,0 +1,9 @@
+coordinates:
+    type: gem
+    provider: rubygems
+    namespace: '-'
+    name: diff-lcs
+revisions:
+    1.6.1:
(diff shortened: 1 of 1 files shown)


capfei:
@btanasie Thank you for the contribution! Can you update your PR to `Artistic-1.0-Perl OR GPL-2.0-or-later OR MIT` and make the spaces 2 instead of 4? We use 2.

The Artistic license used is the Perl version, not 2.0.
 
Artistic Perl 1.0 license text:
https://github.com/halostatue/diff-lcs/blob/v1.6.1/docs/artistic.txt

btanasie:
Thanks you @capfei, I just updated based on the feedback 👍

PR #29591, opened 2025-04-02
Create github-auth.yaml

Pizza-Ria:
Try #2 - a little confused why you don't want to see the full text/link that I used to do my curation - like a proof of the conclusion?

### DIFF
--- a/curations/gem/rubygems/-/github-auth.yaml
+++ b/curations/gem/rubygems/-/github-auth.yaml
@@ -1,0 +1,8 @@
+coordinates:
+  name: github-auth
+  provider: rubygems
+  type: gem
+revisions:
+  3.0.0:
+    licensed:
+      declared: MIT


capfei:
@Pizza-Ria Thank you for the contribution! The proof would be provided in the PR comment details, not the file itself. The file is meant to correct what you see on the site or when the data is used. If you use the site or the API, reason for the change is part of the request. They end up looking like this PR: https://github.com/clearlydefined/curated-data/pull/29590

API example: https://docs.clearlydefined.io/docs/curation/curation-api#complete-json-payload-example

You can do the curation directly on the website, which provides a form to fill out details.
https://docs.clearlydefined.io/docs/curation/curation-process

If you run into an error while signed in it to the ClearlyDefined site (not GitHub), then simply log out of the site and submit again. There is an open issue.

capfei (approved):

Pizza-Ria:
Whoo-hoo!!! Finally, had a chance to come back and check this - fantastic!!!  Thanks for all the guidance on how to set this up and now maybe I can scale it out a bit.

PR #29690, opened 2025-05-21
Add curation for gem oauth2

tangtony:
at https://github.com/oauth-xx/oauth2/blob/main/LICENSE.txt


capfei (approved):

capfei:

PR #29782, opened 2025-07-11
Add curation for go mysql

geramirez:
The correct license is Mozilla Public License Version 2.0 which can be found at https://github.com/go-sql-driver/mysql/blob/master/LICENSE


capfei:
This version has already been curated as MPL-2.0. Closing PR.

PR #29795, opened 2025-07-15
jfractionlab/0.92-2

clearlydefinedbot:
**Type:** Incomplete

**Summary:**
jfractionlab/0.92-2

**Details:**
Found this version in https://packages.debian.org/bullseye/jfractionlab

**Resolution:**
Format: http://dep.debian.net/deps/dep5/
Upstream-Name: JFractionLab
Source: http://sourceforge.net/projects/jfractionlab/

Files: *
Copyright: Copyright (C) 2005-2009 Jochen Georges <gnugeo@gnugeo.de>
License: GPL-3

Files: debian/*
Copyright: Copyright 2009 José L. Redrejo Rodríguez <jredrejo@debi

**Affected definitions**:
- [jfractionlab 0.92-2](https://clearlydefined.io/definitions/debsrc/debian/-/jfractionlab/0.92-2)

PR #29796, opened 2025-07-15
jfractionlab/0.91-3

clearlydefinedbot:
**Type:** Incorrect

**Summary:**
jfractionlab/0.91-3

**Details:**
There are actually 2 licenses and one is not FOSS - see the license file
https://sources.debian.org/src/jfractionlab/0.91-3/LICENCE.txt/

JFractionLab
is free software under the terms of the GNU GPL, Version 3.
(see http://gplv3.fsf.org/
or below)

TableLayout
is free software under the terms of th

**Resolution:**
Fix license - I don't see any "or later" designations on the GPL-3.0 designation.  

**Affected definitions**:
- [jfractionlab 0.91-3](https://clearlydefined.io/definitions/debsrc/debian/-/jfractionlab/0.91-3/0.91-3)


Pizza-Ria:
Not sure what to do about the fact that TWO licenses are in the License.txt file and one appears to not be FOSS (couldn't access the link is dead).

JFractionLab
is free software under the terms of the GNU GPL, Version 3.
(see http://gplv3.fsf.org/
or below)

TableLayout
is free software under the terms of the Clearthought Software License, Version 2.0
(see https://tablelayout.dev.java.net/files/documents/3495/59349/License.txt
or below)

capfei:
@Pizza-Ria I sounds like third-party license so I would leave off the declared. Also, I think this should be "-or-later" per the source file headers.
https://sources.debian.org/src/jfractionlab/0.91-3/src/jfractionlab/ConfManager.java/

What do you think?

Pizza-Ria:
On the third party license - I may need to be better educated on what the goal of this project is - is it ONLY to provide the FOSS Licensing?  Because if I'm trying to figure out what the licensing of a component is and we neglect to provide a clearly listed license then a user may think that the component is ONLY FOSS when there is something else that applies.  That concerns me.

Second, impressed that you found this in Debian - I kept running into dead links!  I agree that the artifacts you found clearly states GPL-3.0-or-later but I went back to the https://jfractionlab.sourceforge.net/ which provides a link to only the GPL-3.0 - http://www.gnu.de/documents/gpl.de.html.  So, this one seems a bit thorny.

Pizza-Ria:
This could also affect - https://github.com/clearlydefined/curated-data/pull/29795 which was closed as I had based the review on the headers found.  I also found this link:  https://sources.debian.org/src/jfractionlab/0.91-3/LICENCE.txt/ which only lists V3 in the license.txt file.

capfei:
@Pizza-Ria When it comes to the A/L/GPL licenses, you have to look at the file headers, usually, for the declaration of "only" or "or-later". The license text used is the same.

Diffing the license text for GPL 3.0 shows this:
image

The SPDX pages for these has the notes at the top explaining that you should check which applies in the file code and has the text commonly used as "Standard License Header" at the bottom of the page each of these.
https://spdx.org/licenses/GPL-3.0-only.html
https://spdx.org/licenses/GPL-3.0-or-later.html

ClearlyDefined is largely used for licensing but also source location and copyright information. The top level `declared` is for the component's license (we support SPDX identifiers and ScanCode's LicenseRef database), that is, what license applies overall. ClearlyDefined also supports doing file level curations to make included third-party code license and copyright information clearer. Most people tend to contribute the main `declared` license for the component. On a ClearlyDefined component page, other licenses found will be listed in the `discovered` licenses so we do make it known that we detected other licenses that may apply. You can also browse the files to see which were detected with other potential license information.

Pizza-Ria:
Yes, the file headers are generally the "source of truth" - I was questioning whether they were actually valid given that the original project seems to be GPL-3.0 (no or later specified).  See also, this attached file from the download of JFractionLab_091 (couldn't find a -3 version of this in SourceForge).  [pack-JFractionLab.zip](https://github.com/user-attachments/files/21354465/pack-JFractionLab.zip).

capfei:
I couldn't open your link but I did download the source zip for `_91` and do see the "or later" in the file headers.
https://sourceforge.net/projects/jfractionlab/files/jfractionlab/jfractionlab_091/JFractionLab_091_src.zip/download

```
/**
 *	Copyright (C) 2005 jochen georges, gnugeo _ at _ gnugeo _ dot _ de
 *
 *	This program is free software: you can redistribute it and/or modify
 *	it under the terms of the GNU General Public License as published by
 *	the Free Software Foundation, either version 3 of the License, or
 *	(at your option) any later version.
 *
 *	This program is distributed in the hope that it will be useful,
 *	but WITHOUT ANY WARRANTY; without even the implied warranty of
 *	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *	GNU General Public License for more details.
 *
 *	You should have received a copy of the GNU General Public License
 *	along with this program.  If not, see <http://www.gnu.org/licenses/>.
 **/
```

I would say that `GPL-3.0-or-later` is probably the correct declared license for jfractionlab 0.91-3.

Pizza-Ria:
@capfei Yes, I agree that the file headers include the "or-later" indicator.  I would think the license.txt file should also indicate the "or-later" but I'm ok with accepting the header designation as sufficient for identification here.  Not sure where we landed on the proprietary license "TableLayout is free software under the terms of the Clearthought Software License, Version 2.0 (see https://tablelayout.dev.java.net/files/documents/3495/59349/License.txt or below).  I'm OK with this not being in the Declared License if you don't want to put non-OSS there but it seems like it should be surfaced SOMEWHERE otherwise people relying on the ClearlyDefined data will be missing part of the picture.

Pizza-Ria:
@capfei Did this one get resolved?  What will happen with the Clearthought Software License?

capfei:
@Pizza-Ria I've been trying to look through the files contained in the source package and I cannot find the TableLayout files. Their website says "The used library "TableLayout.jar" is..." but I cannot find this jar nor files that contain "TableLayout" in the name. Can you point me to them? 
https://clearlydefined.io/definitions/debsrc/debian/-/jfractionlab/0.91-3/0.91-3
https://sources.debian.org/src/jfractionlab/0.91-3

If the TableLayout files are not actually contained within the source package, but in the compiled package, then that license shouldn't be included in this curation.

Pizza-Ria:
See, JFractionLab_091_src\jflsrc_091\libs[TableLayout-bin-jdk1.5-2009-06-10.jar.zip](https://github.com/user-attachments/files/22373544/TableLayout-bin-jdk1.5-2009-06-10.jar.zip)

capfei:
I don't see that zip in https://sources.debian.org/src/jfractionlab/0.91-3. I did come across https://sources.debian.org/src/jfractionlab/0.91-3/debian/README.Debian, which states:
```
jfractionlab for Debian
-----------------------

 The directories 
	/bin
	/libs
	/mkInstaller
	mkJflJar.sh
	icons/jfractionlab_banner_vertical.*
	icons/jfractionlab_icon.eps
	icons/jfractionlab_icon.ico
	icons/splash.xcf
	
 have been removed from upstream sources, as they contained non-free and
 binaries unneded to compile and use this software.	
 -- José L. Redrejo Rodríguez <jredrejo@debian.org>  Fri, 04 Dec 2009 17:44:58 +0100
```

Since I can't find TableLayout in the contents on Debian Sources (https://sources.debian.org/src/jfractionlab/0.91-3), I would say that the above applies and the Clearthought Software License shouldn't be included as part of the declared license in the curation.

Pizza-Ria:
OK, fair enough.  You have convinced me @capfei!

PR #29807, opened 2025-07-21
Add curation for pypi sqlalchemy

SophiaSanles:
at https://docs.sqlalchemy.org/en/20/copyright.html


capfei:
@SophiaSanles The version you are curating is missing. Can you update the PR to include it?

PR #29809, opened 2025-07-21
Update abundance.yaml

Pizza-Ria:
Updated to add quotes to encapsulate special chars in the (c) statement.  
[abundance.zip](https://github.com/user-attachments/files/21355323/abundance.zip)


Pizza-Ria:
That worked - resynced and then added as a new pull request so closing the old one.  Thanks for the hint @capfei!

PR #29822, opened 2025-07-23
Add curation for npm applicationinsights-common

tpope:
at https://github.com/microsoft/ApplicationInsights-JS/blob/main/shared/AppInsightsCommon/LICENSE


capfei:
@tpope Thank you for the updated PRs! They have been merged.

PR #29869, opened 2025-08-14
Updating some OTHER with new SPDX ids

capfei:


capfei:
These are new SPDX licenses added in 3.27.0.

PR #29904, opened 2025-09-12
npmjs/cookie-signature: Declare license to be MIT

reedloden:
For some reason, the `cookie-signature` package is being declared as both MIT and unknown license. It's just MIT. Declare it as such.

https://github.com/tj/node-cookie-signature

https://clearlydefined.io/definitions/npm/npmjs/-/cookie-signature/1.2.2

License: https://github.com/tj/node-cookie-signature/blob/master/LICENSE
Also declared here: https://github.com/tj/node-cookie-signature/blob/b7bd4cb9500bfa5e696143f51d61e5f24f7a625d/package.json#L8

### DIFF
--- a/curations/npm/npmjs/-/cookie-signature.yaml
+++ b/curations/npm/npmjs/-/cookie-signature.yaml
@@ -6,3 +6,30 @@ revisions:
   1.0.1:
     licensed:
       declared: MIT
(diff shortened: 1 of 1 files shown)


capfei:
@reedloden Thank you for the contribution! 🎉

PR #29916, opened 2025-09-23
Revise wcwidth package versions and licenses

abriere-coveo:
re-order versions and add 0.2.14


capfei:
@abriere-coveo Thank you for the contributions! FYI, the ordering we use is alphabetical not numerical. The next time someone curates this package through the service, the PR will auto include the order change back to this.

PR #29934, opened 2025-09-26
Create abiquo-etk.yaml

Pizza-Ria:
Downloaded the archive from https://index.rubygems.org/gems/abiquo-etk/versions/0.4.32 to extract the release date/copyright and license.  Not sure if this is in the correct format for a PR (still find that most confusing).

### DIFF
--- a/curations/gem/rubygems/-/abiquo-etk.yaml
+++ b/curations/gem/rubygems/-/abiquo-etk.yaml
@@ -1,0 +1,37 @@
+```json
+{
+  "contributionInfo": 
+  {
+    "summary": "abiquo-etk 0.4.32 missing",
+    "details": "Add license information, release date and attributions for abiquo-etk 0.4.32",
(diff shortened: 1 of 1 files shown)


capfei:
@Pizza-Ria The json format is used for the API, not curations through GitHub directly. You still need to use the YAML format for those. If you want to try curating using the API, you can give it a try through the our Swagger site: https://api.clearlydefined.io/api-docs/#/. Use `PATCH` in the `curations` section.

Also, I would change the servers (dropdown at the top) to https://dev-api.clearlydefined.io so it generates test pull requests to the [curated-data-dev](https://github.com/clearlydefined/curated-data-dev) repo. Don't include the markdown backticks (```) in the request body.

When you want to submit an actual pull request, change the server back https://api.clearlydefined.io so it will create it in this repo.

I hope that makes sense.

I am closing PR since I can't merge this but do ping me if you need assistance.

PR #32415, opened 2026-04-28
Add curation for pypi tqdm

mnaylor5:
The correct license is MIT AND MPL-2.0 which can be found at https://github.com/tqdm/tqdm/blob/master/LICENCE

### DIFF
--- a/curations/pypi/pypi/-/tqdm.yaml
+++ b/curations/pypi/pypi/-/tqdm.yaml
@@ -1,8 +1,12 @@
 coordinates:
-  name: tqdm
-  provider: pypi
   type: pypi
+  provider: pypi
+  namespace: ""
+  name: tqdm
 revisions:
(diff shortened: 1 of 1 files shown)


capfei:
@mnaylor5 We are not able to do wildcards for versions. Each curation needs to be for a version or commit ID (for source repos). As for the license, we have been curating this as MPL since it is MPL at this point. MIT is listed for the README, _tqdm.py, .gitignore (which isn't in the pypi) and makes more sense to curate those files with MIT instead of the including it as the declared for the project.

We have had discussions about this in the past and point to this discussion where the current maintainer states that it is basically all MPL now: https://github.com/tqdm/tqdm/issues/819.

mnaylor5:
Thanks for clarifying @capfei! I'll close this PR 😄

